PR #7, opened 2018-01-17
Add ability to set a default sort attribute

mpociot:
This PR adds the ability to define a default sort parameter to use when no sort query parameter is available.

It also fixes an issue that occurred when defining allowedSorts, but then perform a request without a sort parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,18 @@ $users = QueryBuilder::for(User::class)
 // Will throw an `InvalidQuery` exception as `password` is not an allowed sorting property
 ```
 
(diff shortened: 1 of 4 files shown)


freekmurze:
Very nice, thanks!

Could you document `defaultSort` in the readme?

mpociot:
Sure - just wanted to know your thoughts on this first.
I'll update the PR tomorrow 😴

mpociot:
@freekmurze Rebased the PR and added the `defaultSort` method to the documentation.

freekmurze:
Thanks!! 🤖

Issue #8, opened 2018-01-18
Exception thrown when no sort is provided in query string and calling allowedSorts


DigitalMachinist:
It seems as though when `allowedSorts()` is called, not including a `sort` in the query string, the sort is assumed to be "" (empty string) and an exception is thrown.

The query string:
```
.../transactions?filter[type]=deposit
```

The exception:
```js
"Exception": {
    "messages": [
        "Given sort `` is not allowed. Allowed sorts are `id, created_at, type, amount, balance`."
    ],
    "exception": "Spatie\\QueryBuilder\\Exceptions\\InvalidQuery",
    "file": "/Users/jeffrose/projects/rivalry/vendor/spatie/laravel-query-builder/src/Exceptions/InvalidQuery.php",
    "line": 27
}
```

The code:
```php
$paginator = QueryBuilder::for($query)
    ->allowedFilters(Filter::exact('id'), 'created_at', Filter::exact('type'), 'amount', 'balance')
    ->allowedSorts('id', 'created_at', 'type', 'amount', 'balance') // If I comment out this line no exception is thrown
    ->paginate($pageSize);
```

**The workaround:**

Leave out the call to `allowedSorts()` and everything's fine.

If I get a chance, I'll submit a PR in the coming days to address this.

PR #9, opened 2018-01-19
Do not handle empty sort query.

pactode:
This is a fix for issue #8 

If nothing is returned from the `$this->request->sort()`, then we return early.


AlexVanderbist:
Perfect! Thanks 👍

PR #11, opened 2018-01-20
Escape LIKE queries

ntzm:
I've made a PR to deal with this in core: laravel/framework#22864

Basically there was inconsistency between some DB servers, and also %, _ and \ weren't being escaped.

### DIFF
--- a/src/Filters/FiltersPartial.php
+++ b/src/Filters/FiltersPartial.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\QueryBuilder\Filters;
 
+use Illuminate\Support\Facades\DB;
 use Illuminate\Database\Eloquent\Builder;
 
 class FiltersPartial implements Filter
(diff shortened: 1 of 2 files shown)


src/Filters/FiltersPartial.php
```diff
@@ -11,11 +12,20 @@ public function __invoke(Builder $query, $value, string $property) : Builder
...
             });
         }
 
-        return $query->where($property, 'LIKE', "%{$value}%");
+        return $query
+            ->where($property, 'LIKE', DB::raw('? ESCAPE "\"'))
+            ->addBinding("%{$this->escapeLike($value)}%");
+    }
+
+    private function escapeLike(string $value): string
```
AlexVanderbist: Can we make this one `protected`? This way it's a bit easier for people to extend if they'd want to.

ntzm: Of course!

AlexVanderbist:
Hi @ntzm 

After merging this in we still ran into some problems. In tests using sqlite everything seems to work perfectly fine.
On MySQL however the `ESCAPE` part of the query seems to be escaped by Laravel itself and ends up looking something like this:
```sql
select * from `media` where `name` LIKE %test% ESCAPE \"\\\"
```

Any ideas for a quick fix?

ntzm:
Damn sorry, should have tested. I'm at work now, I'll figure something out later. Probably worth reverting if you have more stuff to release. My bad

AlexVanderbist:
Don't worry about it! Reverted it for now but I'll keep an eye on laravel/framework#22864 for any updates as well 👍

ntzm:
Thanks, and sorry again!

ntzm:
I can't for the life of me figure out how to prevent that from happening

PR #15, opened 2018-01-28
Added translations for exception messages

shabushabu:
I need this for an app I'm working on and thought it might be useful for others. Only thing I did was move the lang strings to a translation file and added german translations. Would be great if you could merge this! Cheers and keep up the great work!

### DIFF
--- a/lang/de/errors.php
+++ b/lang/de/errors.php
@@ -1,0 +1,7 @@
+<?php
+
+return [
+    'filters'  => 'Angegebene Filter `:unknown` sind nicht erlaubt. Erlaubte Filter sind `:allowed`.',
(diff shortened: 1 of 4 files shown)


AlexVanderbist:
Hi @shabushabu, thanks for your PR! Different locale are an important use-case which I definitely overlooked when adding the exceptions. However we'd rather not maintain exception messages in all different languages in this package.

I've opened #18 which introduces specific exceptions per invalid query parameter. This way you could catch these exceptions in your own exception handler and return an appropriate HTTP response (localized string or even a JSON object)

Hopefully this works out for your use-case as well :)

shabushabu:
Sure, cheers!

PR #18, opened 2018-01-30
Add specific exceptions for every invalid query

AlexVanderbist:
This PR adds a couple more specific exceptions to be able to better handle different query exceptions in userland.

- `InvalidFilterQuery` with `getAllowedFilters()` and `getUnknownFilters()`
- `InvalidIncludeQuery` with `getAllowedIncludes()` and `getUnknownIncludes()`
- `InvalidSortQuery` with `getAllowedSorts()` and `getUnknownSort()`

All these exceptions extend from `InvalidQuery`.

### DIFF
--- a/src/Exceptions/InvalidFilterQuery.php
+++ b/src/Exceptions/InvalidFilterQuery.php
@@ -8,10 +8,10 @@ use Illuminate\Support\Collection;
 class InvalidFilterQuery extends InvalidQuery
 {
     /** @var \Illuminate\Support\Collection */
-    private $unknownFilters;
+    public $unknownFilters;
 
     /** @var \Illuminate\Support\Collection */
-    private $allowedFilters;
+    public $allowedFilters;
 
     public function __construct(Collection $unknownFilters, Collection $allowedFilters)
     {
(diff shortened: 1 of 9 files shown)


src/Exceptions/InvalidFilterQuery.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public static function filtersNotAllowed(Collection $unknownFilters, Collection $allowedFilters)
+    {
+        return new static(...func_get_args());
+    }
+
+    public function getUnknownFilters(): Collection
+    {
+        return $this->unknownFilters;
```
freekmurze: Let's just make the properties public

freekmurze:
Could you rebase this against the current master?

AlexVanderbist:
Rebased against master 👍

src/Exceptions/InvalidFilterQuery.php
```diff
@@ -0,0 +1,32 @@
...
+    public $allowedFilters;
+
+    public function __construct(Collection $unknownFilters, Collection $allowedFilters)
+    {
+        $this->unknownFilters = $unknownFilters;
+        $this->allowedFilters = $allowedFilters;
+
+        $unknownFilters = $this->unknownFilters->implode(', ');
+        $allowedFilters = $this->allowedFilters->implode(', ');
+        $message = "Given filter(s) `{$unknownFilters}` are not allowed. Allowed filters are `{$allowedFilters}`.";
```
freekmurze: Use `filter(s)` instead of `filters` to be more consistent. Do it also for the other exceptions.

freekmurze:
👍

MatanYadaev:
Hey guys, any idea how to prevent from `InvalidFilterQuery` to be thrown?

keizah7:
@MatanYadaev you cant turn off it in config file
```
    'disable_invalid_filter_query_exception' => true,
```

for some reason I can't it don't work with `InvalidSortQuery`

PR #22, opened 2018-02-01
Add tests and docs for loading nested relations

### LINKED ISSUE #21: Ability to include nested relationships
I would like to be able to include nested relationships like you do with `load` or `with`: `->load('relation1.relation2')`

Obviously this is a bit more difficult because PHP translates `.` to `_` in GET request values

ntzm:
Closes #21

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,17 @@ $users = QueryBuilder::for(User::class)
 // $users will contain all users with their posts and permissions loaded
 ```
 
+You can load nested relationships using `.`:
+
+``` php
+// GET /users?include=posts->comments,permissions
+$users = QueryBuilder::for(User::class)
+    ->allowedIncludes('posts, 'posts.comments', 'permissions')
+    ->get();
+
+// $users will contain all users with their posts, comments on their posts and permissions loaded
+```
+
 When trying to include relationships that have not been allowed using `allowedIncludes()` an `InvalidQuery` exception will be thrown.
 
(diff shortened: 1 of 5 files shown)


AlexVanderbist:
Thanks for the PR! If at all possible I'd prefer to use the dot-notation to adhere to the JSON API spec: http://jsonapi.org/format/#fetching-includes

ntzm:
I'm not sure if that's possible with PHP, as it converts `.` to `_` in GET parameters, I don't know if there's a way of forcing Laravel's request instance to parse `$_SERVER['QUERY_STRING']` instead of `$_GET`, as that would work. I'll look into it

ntzm:
I've updated it to use `.` instead, but this could be a BC for anyone using `_` instead of `-` in relation names. This is undocumented behaviour but does happen due to using `camel_case`

jshayes:
@ntzm From my testing, it seems like PHP replaces `.` with `_` for the query key, not the value. So when I do `?include=foo.bar` the `.` is not replaced with `_`. I get the following
```
array:1 [
  "include" => "foo.bar"
]
```
But when I do `?foo.bar`, I get
```
array:1 [
  "foo_bar" => ""
]
```

ntzm:
You're right @jshayes, updating

ntzm:
So you could already do it 🤦‍♂️ , turning this PR into docs and tests update

freekmurze:
Thanks 👍

PR #35, opened 2018-02-08
Added support for selecting columns.

cornelp:


AlexVanderbist:
Hi, thanks a lot for the PR! I like the idea and the code looks good. 

Support for selecting specific columns is defined in the spec as sparse fieldsets: http://jsonapi.org/format/#fetching-sparse-fieldsets. If we can make this PR use the same parameters as the spec (`GET /some-model?fields[some-model]=name&fields[related-model]=id,name`) I'll merge it in :)

AlexVanderbist:
Thanks!

PR #37, opened 2018-02-09
Fix error when filter parameter is array

spyric:
Hello.

In JSON API Spec there is no requirement that filter parameter should be string. So It's ok when user will pass something like `?filter[column_name][max]`, but currently in that case exception will be thrown. 

This PR make is possible to use arrays in filters.


freekmurze:
Thanks!

PR #39, opened 2018-02-13
Add allowedSearches to QueryBuilder.

pactode:
Added the possibility to easily search in multiple columns by the `q` query parameter.

I know this sort of breaks the JSON API standard (nothing documented in relation to searching), but I like having this feature in my API's. If you like it I can eventually implement searching in relationship columns and probably more useful features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,26 @@ $users = QueryBuilder::for(User::class)
 // $users will be sorted by name in ascending order with a secondary sort on street in descending order.
 ```
 
+### Searching
+
+The `q` query parameter is used to search in your data by the defined columns in allowedSearches.
+
+You can define multiple columns to search in by separating them with a comma.
+
+```php
+QueryBuilder::for(User::class)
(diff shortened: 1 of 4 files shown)


AlexVanderbist:
Thanks a lot for the PR! Not gonna merge this in for now as it's not part of the JSON API spec. However I really like the idea of easily filtering on partial matches for multiple columns. 

You can already achieve this using custom filters right now:
```php
->allowedFilters(Filter::custom('q', SearchFilter::class))
// GET /endoint?filter[q]=search+query
```
The only issue here is you'd have to define in the fields in the `SearchFilter` class. Maybe we can allow  instances of custom filter classes to be passed to the `Filter::custom` method. That way something like this would be possible:
```php
->allowedFilters(Filter::custom('q', new SearchFilter('first_name', 'last_name)))
```

pactode:
That would definitely be the better option. I like having the flexibility to pass in the columns dynamically instead of being forced to define a search filter for each model.

PR #47, opened 2018-03-21
Allow setting custom parameter names using query-builder config

vyuldashev:
This PR will allow using custom parameter names using config.

For example, if we want to change `filter` parameter to `filters` we will be able to achieve it using configuration. 

Publishing configuration file is optional.

### DIFF
--- a/config/query-builder.php
+++ b/config/query-builder.php
@@ -2,6 +2,9 @@
 
 return [
 
+    /**
+     * Define query parameter names used in the query.
+     */
     'parameters' => [
         'include' => 'include',
 
--- a/config/query-builder.php
+++ b/config/query-builder.php
@@ -1,0 +1,15 @@
+<?php
+
+return [
+
+    'parameters' => [
+        'include' => 'include',
+
+        'filter' => 'filter',
+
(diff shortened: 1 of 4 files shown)


config/query-builder.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+return [
+
```
freekmurze: Add a few sentences here that explain what this option does

config/query-builder.php
```diff
@@ -0,0 +1,15 @@
...
+        'include' => 'include',
+
+        'filter' => 'filter',
+
+        'sort' => 'sort',
+
+        'fields' => 'fields',
+    ],
+
+];
```
freekmurze: Add a newline here

config/query-builder.php
vyuldashev: Added, by need help with finding more words 😄

freekmurze:
Could you also add tests to prove that this works?

vyuldashev:
@freekmurze Added tests.

freekmurze:
Thanks!

PR #50, opened 2018-03-27
Support allowedAppends()

robsontenorio:
# Append attributes

Sometimes you will want to append some custom attributes into result from a Model. This can be done using the `append` parameter.

```
class User extends Model{

    getFullnameAttribute()
    {
        return $this->firstname.' '.$this->lastname;
    }
}
```

```
// GET /users?append=fullname

    ->allowedAppends('fullname')
    ->get();
```

Of course you can pass a list of attributes to be appended.

```
// GET /users?append=fullname,ranking
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,6 +308,35 @@ GET /posts?include=author&fields[author]=name
 
 All posts will be fetched including only the name of the author. 
 
+### Append attributes
+
+Sometimes you will want to append some custom attributes into result from a Model. This can be done using the `append` parameter.
+
+```
+class User extends Model{
+
+    getFullnameAttribute()
+    {
+        return $this->firstname.' '.$this->lastname;
+    }
+}
+```
+
+```
+// GET /users?append=fullname
+
+$users = QueryBuilder::for(User::class)
+    ->allowedAppends('fullname')
+    ->get();
+```
+
+Of course you can pass a list of attributes to be appended.
+
+```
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -308,6 +308,35 @@ GET /posts?include=author&fields[author]=name
...
 All posts will be fetched including only the name of the author. 
 
+### Append attributes
+
+Sometimes you will want to append some custom attributes into result from a Model. This can be done using the `append` parameter.
+
+```
+class User extends Model{
+
+    getFullnameAttribute()
```
freekmurze: Add the `function` keyword here

src/Exceptions/InvalidAppendQuery.php
```diff
@@ -0,0 +1,32 @@
...
+class InvalidAppendQuery extends InvalidQuery
+{
+    /** @var \Illuminate\Support\Collection */
+    public $unknownAppends;
+
+    /** @var \Illuminate\Support\Collection */
+    public $allowedAppends;
+
+    public function __construct(Collection $unknownAppends, Collection $allowedAppends)
+    {
```
freekmurze: Let's do all this in the `appendsNotAllowed` function.

robsontenorio:
@freekmurze Done!

freekmurze:
Thank you!

Issue #54, opened 2018-04-18
Nested relation includes too strict?


slashequip:
Hey,

Given this query builder:
```
QueryBuilder::for(Site::class)
    ->allowedIncludes('division.organisation')
    ->get();
```

If I only try to include `division` in my request I get an error saying `division` isn't in the list of allowed includes. Is this intended behaviour? Feels too restrictive/strict to me.

The work around is to do this:
```
QueryBuilder::for(Site::class)
    ->allowedIncludes('division', 'division.organisation')
    ->get();
```

Feels a little repetitive?

If I can get some time I don't mind attempting a PR but let me know how you feel about this.

freekmurze:
I'd accept a PR with tests.

PR #55, opened 2018-04-18
No longer requires explicit defining of parent includes in nested queries.

slashequip:
PR to cover #54.

Before:
```
// GET /users?include=posts
QueryBuilder::for(User::class)
    ->allowedIncludes('posts', 'posts.comments')
    ->get();
```

After:
```
// GET /users?include=posts
QueryBuilder::for(User::class)
    ->allowedIncludes('posts.comments')
    ->get();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ You can load nested relationships using `.`:
 ``` php
 // GET /users?include=posts.comments,permissions
 $users = QueryBuilder::for(User::class)
-    ->allowedIncludes('posts', 'posts.comments', 'permissions')
+    ->allowedIncludes('posts.comments', 'permissions')
     ->get();
 
 // $users will contain all users with their posts, comments on their posts and permissions loaded
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -144,7 +144,7 @@ class QueryBuilder extends Builder
                 return collect(explode('.', $include))
                     ->reduce(function ($collection, $include) {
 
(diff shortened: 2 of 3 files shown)


AlexVanderbist:
Hi, thanks a lot for the PR! 

I feel like allowing includes like this is not always intended behaviour. You might for example have an endpoint to get a post and its author: `GET /posts/123?include=author`. However you don't want to allow for queries like `GET /posts/123?include=author.posts` as that'll expose all other posts by the author.

Would it be possible to allow adding nested includes by using a wildcard?
E.g. `->allowedFilters('user.*')` would allow including users and all the user's relationships.

slashequip:
@AlexVanderbist no problem!

I think we have crossed wires on what this PR does! This just stops you having to define all the parent relationships separately & explicitly.

Say I have this `->allowIncludes('posts.comments')` on a QueryBuilder.

If I tried to make this request `GET /author/123?include=posts` it would currently throw an error as you haven't explicitly defined allowing the `posts` relationship.

You would have had to do `->allowIncludes('posts', 'posts.comments')` to not get an error. In this PR when you do `->allowIncludes('posts.comments')` it basically automagically does this `->allowIncludes('posts', 'posts.comments')` for you.


This PR *DOES NOT* auto allow further child model relationships! Hope we are on the same page now 😁 

P.S I do like the idea of wildcard though but I guess it would require some from of model/method sniffing or lots of try/catching.

slashequip:
Phew! No problem 👍

PR #59, opened 2018-04-25
Add Filter for Scope

gpibarra:


freekmurze:
Could you add a test and add an example to the readme?

gpibarra:
In the model there is a Scope, then the filter calls the Scope where the entire logic of the filter is found

```
//Controller
        $sales = QueryBuilder::for(Sale)
            ->allowedFilters([
                Filter::scope('DateFrom'),
                Filter::scope('DateUntil'),
                ])
            ->get();

```
```
//Model
    public function scopeDateFrom(Builder $query, date) :Builder
    {
        $objDate = Carbon::parse(date);
        return $query->where('DATETIME','>=',$objDate>format('y-m-d'));
    }
```

freekmurze:
Could you PR that example to the readme?

Issue #36, opened 2018-02-09
Sorting by related properties


AlexVanderbist:
As described in the spec here: http://jsonapi.org/format/#fetching-sorting

>It is recommended that dot-separated (U+002E FULL-STOP, “.”) sort fields be used to request sorting based upon relationship attributes. For example, a sort field of author.name could be used to request that the primary data be sorted based upon the name attribute of the author relationship.

However this is not possible on an eloquent query builder. `Model::with('related')->orderBy('related.name')->get()` doesn't work as the related models would be fetched in a second query and the first query doesn't know the `related.name` field. 
I'm also not sure what the intended behaviour would be relationships returning multiple models. (Sort the related models first; then sort parent models by first related model?)

As I see it there a couple different solutions: 
- if there are any sort queries for related properties (containing `.`) fetch the results first and sort the collection after that. This doesn't work if you're paginating or limiting the query
- use a select to include the related property on the initial query and then sort – only works for `belongsTo` and `hasOne` relationships and feels like a pain to implement

Any thoughts/insights on this are very welcome!

ntzm:
Would be a great feature but I'm not sure how possible it is without joins. I tried to do something similar on a personal project and ended up just abandoning the idea because I couldn't find a reliable way to get it to work for every relationship type.

brendt:
> but I'm not sure how possible it is without joins

Is there a problem to use joins?

Another thing coming to mind, my first thought when I was trying to do this without knowing it wasn't possible: I wouldn't mind implementing this sort myself if that makes it easier for the package, something like:

```php
$favorites = QueryBuilder::for(Favorite::class)
    ->customSorts([
        'loan_type_name' => function (Builder $builder, bool $isDesc): Builder {
            return $builder
                ->join('loan_participations', 'favorites.loan_participation_id', '=', 'loan_participations.id')
                ->join('loan_types', 'loan_participations.loan_type_id', '=', 'loan_types.id')
                ->orderBy('loan_types.name', $isDesc ? 'desc' : 'asc');
        }
    ])

    // ...
```

Don't know if this would be possible, I haven't really looked at the internals of the package yet.

robsontenorio:
@AlexVanderbist 
Maybe? https://github.com/fico7489/laravel-eloquent-join

golubkovden:
Yeah, it would be great to see this feature.

matthias-schmitt:
Hello @ all

First of all thanks for the great package!

[...]

PR #64, opened 2018-05-12
Prepend fields with table name in database query

shawm11:
The resulting query from `QueryBuilder` did not work with database joins. Now joins can be used with the resulting database query. This should help with issue #36.


freekmurze:
Thank you!

Issue #65, opened 2018-05-18
order by created_at desc, created_at desc, created_at desc


alexgiuvara:
```php
->defaultSort('-updated_at')
->allowedSorts('id', 'created_at', 'updated_at')
```
Request: `sort=-created_at&page[size]=25&page[number]=1&filter[status]=deleted`
The generated query:
```sql
select * from `posts` where `created_by` in (?) 
and `status` = ? 
order by `created_at` desc, `created_at` desc, `created_at` desc
limit 25 offset 0
```

While debugging, I see that `addSortsToQuery` is called 3 times with this payload:
```php
Collection {#434
  #items: array:1 [
    0 => "-created_at"
  ]
}
```

alexgiuvara:
I changed these two tests on `SortTest.php` to prove what I'm saying:
```php
/** @test */
    public function it_can_allow_multiple_sort_parameters()
    {
        \DB::enableQueryLog();
        $sortedModels = $this
            ->createQueryFromSortRequest('name')
            ->allowedSorts('id', 'name')
            ->get();

        $this->assertSame(\DB::getQueryLog()[0]['query'],'select "test_models".* from "test_models" order by "name" asc, "name" asc');
        $this->assertSortedAscending($sortedModels, 'name');
    }
```

```php
/** @test */
    public function it_can_sort_by_multiple_columns()
    {
        factory(TestModel::class, 3)->create(['name' => 'foo']);
        \DB::enableQueryLog();

        $sortedModels = $this
            ->createQueryFromSortRequest('name,-id')
            ->allowedSorts('name', 'id')
            ->get();

        $expected = TestModel::orderBy('name')->orderByDesc('id');
       
        $this->assertSame(\DB::getQueryLog()[0]['query'],'select "test_models".* from "test_models" order by "name" asc, "id" desc, "name" asc, "id" desc');
        $this->assertEquals($expected->pluck('id'), $sortedModels->pluck('id'));
    }
```
both went green

PR #67, opened 2018-05-19
filter out orderBy duplicates

alexgiuvara:
This corrects the behavior specified here https://github.com/spatie/laravel-query-builder/issues/65
It removes `order by` duplicates for same column, same direction


freekmurze:
Thanks!

PR #70, opened 2018-06-04
Resolve #14 (query with case-insensitive)

### LINKED ISSUE #14: Add ILike varaint for postgres
In mysql "like" is case-insensitive, for postgres ilike is more useful;
or common variant  for any db - "like" compare with LOWER() wrapper for column value and searchable value

ibrunotome:


freekmurze:
Thanks!

PR #72, opened 2018-06-08
Replace 'strtolower' for 'mb_strtolower'

joaorobertopb:
Introduced in #70 

Using `lower()` in database:

```sql
#MySQL
SELECT LOWER(`SÃO PAULO`) FROM DUAL # são paulo
```

Using `strtolower()` in code:

```php
echo strtolower('SÃO PAULO'); # sÃo paulo
```

Using **mb_strtolower()** in code:

```php
echo mb_strtolower('SÃO PAULO', 'UTF8'); # são paulo
```


freekmurze:
Thanks!

PR #74, opened 2018-06-11
Use general grammar wrap method in FiltersPartial

### LINKED ISSUE #73: PostgreSQL exception
#71  seems to break the `FilterPartial` in combination with Postgres.

I am getting:
``
SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: ` character varying
LINE 1: ..._accounts"."user_id" is not null and LOWER(`some_filter`...
``

``
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts. (SQL: select "my_accounts".* from "my_accounts" where "my_accounts"."user_id" = 1 and "my_accounts"."user_id" is not null and LOWER(`some_filter`) LIKE %something%)
``

PostgreSQL 9.6.3

Downgrading to version 1.9.2 solves the issue.

josiasmontag:
Fixes #73

### DIFF
--- a/src/Filters/FiltersPartial.php
+++ b/src/Filters/FiltersPartial.php
@@ -8,7 +8,9 @@ class FiltersPartial implements Filter
 {
     public function __invoke(Builder $query, $value, string $property): Builder
     {
-        $sql = "LOWER(`{$property}`) LIKE ?";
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

Issue #82, opened 2018-07-15
Pagination query parameter page throws an error


rajasekar-d:
Removed filter name from the query parameter by specifying null reference in the config file.

`
'filter' => null,
`

After removing  filter query parameter name from the URL. Now it throws throws an error  when i use the pagination query parameter page **Given filter(s) `page` are not allowed.** 

**Query:**
`
$users = QueryBuilder::for(User::class)->allowedFilters(Filter::custom('role', UserRole::class),Filter::custom('search', UserSearch::class))->with('roles')->paginate(1);
`

**URL**
`?search=john&role=4&page=2`

enkota:
Is there any advancement on this feature? Tried editing the QueryBuilderServiceProvider.php filter but look like this has moved to the BuilderRequest file instead. 

For my API I'd prefer to remove the `filter` requirement but in turn, stops pagination working.

PR #89, opened 2018-08-01
exclude filter parameter config for #82

rajasekar-d:


freekmurze:
Thank you for this PR! Could you add tests that prove that this change works? Please also update the readme with this new functionality.

freekmurze:
Closing due to lack of activity.

PR #92, opened 2018-08-20
Publish config file in console mode

ndnam90:
I use this repository for Lumen app, but config_path() helper does not exists in Lumen.


freekmurze:
Thanks!

PR #103, opened 2018-10-01
Add allowedFields method

### LINKED ISSUE #81: By default sorting by a non-existant property throws an error
If you visit http://my-api.com/users?sort=thisfielddoesnotexist you recieve an sql error:

"SQLSTATE[42S22]: Column not found: 1054 Unknown column 'thisfielddoesnotexist' in 'order clause' (SQL: select `users`.* from `users` order by `thisfielddoesnotexist` asc limit 15 offset 0)"

I didn't set any  explicitly allowed sorting but it seems weird that the default allows for an SQL error to be triggered by the user.

Te7a-Houdini:
this **PR** , adds **allowedFields** method which will thorw **InvalidFieldQuery Exception** if the request has not allowed field .

also i think this **PR** may resolve #81  

``` php
// GET /users?fields[users]=id
    ->allowedFields('name')
    ->get();
 // Will throw an `InvalidFieldQuery` exception as `id` is not an allowed
```

### DIFF
--- a/src/Exceptions/InvalidFieldsQuery.php
+++ b/src/Exceptions/InvalidFieldsQuery.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Spatie\QueryBuilder\Exceptions;
+
+use Illuminate\Http\Response;
+use Illuminate\Support\Collection;
+
+class InvalidFieldsQuery extends InvalidQuery
+{
+    /** @var \Illuminate\Support\Collection */
+    public $unknownFields;
+
+    /** @var \Illuminate\Support\Collection */
+    public $allowedFields;
+
+    public function __construct(Collection $unknownFields, Collection $allowedFields)
+    {
+        $this->unknownFields = $unknownFields;
+        $this->allowedFields = $allowedFields;
+
+        $unknownFields = $unknownFields->implode(', ');
+        $allowedFields = $allowedFields->implode(', ');
+        $message = "Given field(s) `{$unknownFields}` are not allowed. Allowed field(s) are `{$allowedFields}`.";
+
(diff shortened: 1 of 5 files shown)


src/Exceptions/InvalidFieldsQuery.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Spatie\QueryBuilder\Exceptions;
+
+use Illuminate\Http\Response;
+use Illuminate\Support\Collection;
+
+class InvalidFieldsQuery extends InvalidQuery
```
AlexVanderbist: Can we rename this class to `InvalidFieldQuery`? That's a bit more in line with the other exceptions :)

src/QueryBuilder.php
```diff
@@ -108,6 +112,17 @@ public function allowedFilters($filters) : self
         return $this;
     }
 
+    public function allowedFields($fields) : self
```
AlexVanderbist: Can we add support for `$builder->allowedFields('*')` and enable that by default? 

A bit like `$builder->allowedSorts('*')` - this way this entire feature will be backward compatible!

Te7a-Houdini: @AlexVanderbist i added support for ```$builder->allowedFields('*')``` but i think by default the method allows all fields if nobody used ```$builder->allowedFields(['column'])```  [check this test ](https://github.com/spatie/laravel-query-builder/blob/master/tests/FieldsTest.php#L25)

AlexVanderbist: Thanks, you're right!

AlexVanderbist:
Thanks!

PR #116, opened 2018-10-19
Fix incorrect example about scope filter

oanhnn:
I think `starts_at` before `$date` then `start_at <= $date`.


freekmurze:
Thanks!

PR #118, opened 2018-10-21
Fix corrupted data in 'allowedIncludes'

rainerkent:
Fixes issue #112

### DIFF
--- a/tests/IncludeTest.php
+++ b/tests/IncludeTest.php
@@ -24,6 +24,11 @@ class IncludeTest extends TestCase
             $model
                 ->relatedModels()->create(['name' => 'Test'])
                 ->nestedRelatedModels()->create(['name' => 'Test']);
+
+            $model->relatedThroughPivotModels()->create([
+                'id'   => $model->id + 1,
+                'name' => 'Test'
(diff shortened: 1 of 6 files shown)


tests/IncludeTest.php
```diff
@@ -24,6 +24,11 @@ public function setUp()
             $model
                 ->relatedModels()->create(['name' => 'Test'])
                 ->nestedRelatedModels()->create(['name' => 'Test']);
+
+            $model->relatedThroughPivotModels()->create([
+                'id'   => $model->id + 1,
```
AlexVanderbist: Nice!

AlexVanderbist:
Looks good, thanks for the PR!

rusowyler:
Hi, this solution seems to break the morph eager queries. As it was before it worked correctly for morph relations!

May be some like this could work:

```php
return [$fullRelationName => function ($query) use ($fields) {
    if(!($query instanceOf MorphTo)) {
        $query->select($this->prependFieldsWithTableName($fields, $query->getModel()->getTable()));
    }
}];
```

Or may be I got it all wrong! :(

PR #119, opened 2018-10-23
Enhancement TravisCI

oanhnn:
Enhancement TravisCI:
- Using cache directory `$HOME/.composer/cache` for speed up running
- Add more running environments


freekmurze:
Thanks!

PR #126, opened 2018-10-30
Allow ignorable values

Propaganistas:
Quick & dirty take on #104 .
Proof of concept.

Please adjust style & add tests yourself if this gets merged.

Issue #104, opened 2018-10-02
Provide possibility to ignore filters on null value


Propaganistas:
```php
'filter' => [
    'name' => 'John',
    'email' => null,
]
```

Current behavior: searching for models with (partial) name "John" AND an empty email address.
Expected behavior: searching for models with (partial) name "John". Email address doesn't matter since it isn't being filtered.

Or at least some option should be provided to ignore `null`...

michgeek:
If you don't want to do the manual filtering you can put it in a middleware wherever needed

```php
class RemoveParametersWithEmptyValueFromQueryString {

    public function handle($request, Closure $next)
    {
        $request->query->replace(array_filter($request->query->all(), 'mb_strlen'));
        return $next($request);
    }

}
```

michgeek:
If you don't want to manually filter everywhere you can put it in a middleware. Since filters are only 1 level deep you can loop on the first level. If you need to filter more deeply see the second solution

Solution 1: only first level

```php
class RemoveParametersWithEmptyValueFromQueryString {

  /**
   * Remove parameters with empty value from query string
   *
   * @param $request
   * @param Closure $next
   *
   * @return mixed
   */
  public function handle($request, Closure $next) {
    $query = $request->query->all();

    foreach($query as $key => $values) {
      $query[$key] = array_filter($values, 'strlen');

      if(empty($query[$key])) {
        unset($query[$key]);
      }
    }

    $request->query->replace($query);

    return $next($request);
  }
}
```

Solution 2: deeply nested filters

```php
class RemoveParametersWithEmptyValueFromQueryString {

  /**
   * Remove parameters with empty value from query string
   *
   * @param $request
   * @param Closure $next
   *
   * @return mixed
   */
  public function handle($request, Closure $next) {
    $query = $request->query->all();
    $request->query->replace($this->remove_empty_parameters($query));

    return $next($request);
  }

  /**
   * Remove empty parameters
   *
   * @param $array
   *
   * @return mixed
   */
  public function remove_empty_parameters($array) {
    foreach ($array as $key => &$value) {
      if (is_array($value)) {
        $value = $this->remove_empty_parameters($value);

        if (!sizeof($value)) {
          unset($array[$key]);
        }
      } elseif (!strlen($value)) {
        unset($array[$key]);
      }
    }

    return $array;
  }

}
```

hulkur:
[...]

PR #131, opened 2018-11-20
Allow for ignorable values on filters

dominikb:
This addresses  #104 and is inspired by #126 | Thanks @Propaganistas

## Changes to original implementation
* When the filter gets called with an array of values, the non-ignored subset propagates to the QueryBuilder
* Signature of `ignore()` allows for more flexible calls

The implementation makes use of `Illuminate\Support\Collection` for easier readability.
If requested, I can, of course, switch the implementation to use native array functions only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,36 @@ $users = QueryBuilder::for(User::class)
     ->get();
 // $users will contain all users that have the `createPosts` permission
 ```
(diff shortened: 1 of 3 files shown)


AlexVanderbist:
Thanks, this looks really good!

PR #124, opened 2018-10-26
Allow to filter by relation model properties

jl9404:
resolves #101 

```php
$users = QueryBuilder::for(User::class)
    ->allowedFilters(['name', 'related-models.field_name'])
    ->get();
```

Issue #101, opened 2018-09-26
Add support to filter by relationships


chanjsq:
Say filter resource A by specify its related resource B's conditions.

AlexVanderbist:
I would merge a tested PR that adds this. 

#124 is a pretty good first attempt - however it only works for filters with one level of nesting (e.g. `?filter[address.street_name]` but not `?filter[address.country.name]`).

For anyone needing this now: you can easily accomplish the same behaviour using a [scope filter](https://github.com/spatie/laravel-query-builder#scope-filters) or custom filter.

For example using a scope filter:
```php
// in your model
public function streetNameLike($query, $value) {
    return $query->whereHas('address', function ($q) use ($value) {
        return $q->where('street_name', 'LIKE', "%{$value}%")
    });
}
```

```php
QueryBuilder::for(User::class)->allowedFilters([
    Filter::scope('street_name_like'),
])
```

SamDevBR:
@AlexVanderbist  actually I was successful using this package in lumen without the fork, I just had to create two helpers (request, config_path) and publish the service provider and the config file, maybe remove these helper dependicies and create functions to do same would be great, the only thing left to do to use this with Lumen would be create the vendor:publish command, because lumen doesn't have it.

UnnitMetaliya:
seems like filtering on multiple level nesting is possible now. 

Just tried something like: `filter[referralsPromoCode.coupon.type]` and it worked fine.

PR #136, opened 2018-12-07
Re-open: Allow to filter by relation model properties

NickToony:
Re-opening MR by **jl9404**

I am now using this in a project and it is working as intended. All tests are passing in my environment. If any further changes are required, I can look into it. Otherwise, should consider merging this again now that the blocking Laravel bug is resolved.

**Original pull request:**
#124 Allow to filter by relation model properties

**Resolved tickets:**
#101 Add support to filter by relationships

**Example (by jl9404)**
```
    ->allowedFilters(['name', 'related-models.field_name'])
    ->get();
```

### DIFF
--- a/src/Filters/FiltersExact.php
+++ b/src/Filters/FiltersExact.php
@@ -2,16 +2,46 @@
 
 namespace Spatie\QueryBuilder\Filters;
 
+use Illuminate\Support\Str;
+use Illuminate\Support\Collection;
 use Illuminate\Database\Eloquent\Builder;
 
 class FiltersExact implements Filter
 {
+    protected $relationConstraints = [];
+
     public function __invoke(Builder $query, $value, string $property) : Builder
     {
+        if ($this->isRelationProperty($property)) {
+            return $this->withRelationConstraint($query, $value, $property);
+        }
+
         if (is_array($value)) {
             return $query->whereIn($property, $value);
         }
 
-        return $query->where($property, $value);
+        return $query->where($property, '=', $value);
+    }
+
+    protected function isRelationProperty(string $property) : bool
+    {
+        return Str::contains($property, '.') && ! in_array($property, $this->relationConstraints);
+    }
+
+    protected function withRelationConstraint(Builder $query, $value, string $property) : Builder
+    {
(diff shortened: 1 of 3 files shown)


src/Filters/FiltersExact.php
```diff
@@ -2,16 +2,46 @@
...
+                return [
+                    $parts->except(count($parts) - 1)->map('camel_case')->implode('.'),
+                    $parts->last(),
+                ];
+            });
+
+        return $query->whereHas($relation, function (Builder $query) use ($value, $relation, $property) {
+            $this->relationConstraints[] = $property = $query->getModel()->getTable().'.'.$property;
+
+            $this->__invoke($query, $value, $property);
```
AlexVanderbist: Clever!

src/Filters/FiltersPartial.php
```diff
@@ -4,10 +4,14 @@
 
 use Illuminate\Database\Eloquent\Builder;
 
-class FiltersPartial implements Filter
+class FiltersPartial extends FiltersExact implements Filter
```
AlexVanderbist: Making `FiltersPartial` extend `FiltersExact` feels a bit dirty as not every partial filter is an exact filter...
I'll probably refactor this to make `FiltersPartials` and `FiltersExact` both extend a base `Filter` class or use a `FiltersRelatedModels` trait.

AlexVanderbist:
Thanks!

NickToony:
@jelmervisser I'll have some time to look at this tomorrow hopefully - Ideally you shouldn't have to change anything. Sorry for breaking something!

dominikb:
I just opened a pull request (#143) that should fix your issue @jelmervisser.

As far as I could tell the problem was with detecting if the filter should be applied to a relation or not.
By prefixing your filters with the model's table name, it presumed that the filter was for a relation of your `Accommodation::class`

I tried to write the test to explicitly expect the method __testModels()__ not to be called on the `TestModel::class`, but could not find a way to do that because the Eloquent Builder always creates a new instance of the class for building its query.
Does anybody know how to properly test this?

PR #140, opened 2018-12-11
Prepend sorts for selected table name when ambiguous with join

shawm11:
Now sorting by any column will work if a JOIN is used with resulting query. Before, there would be an error from the database about ambiguous ORDER BY column if sorting by a column that is in more than one of the JOINed tables.

### DIFF
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -40,6 +40,12 @@ class QueryBuilder extends Builder
     /** @var \Illuminate\Http\Request */
     protected $request;
 
+    /** @var string */
+    protected $modelTableName;
+
+    /** @var array|null */
+    private $joinColumnConflicts = null;
+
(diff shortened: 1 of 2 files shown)


shawm11:
I changed it so that only a sort columns that are prepended are the ones that would have a collision with one or more JOINed tables. However, the solution only works if the JOINs are in query given to the QueryBuilder::for() function instead of being added to the resulting query.

AlexVanderbist:
Thanks for adding those changes. I still feel like it's all a bit too brittle to merge right now.

Once #164 gets merged I think you should be able to do something like this;
```php
QueryBuilder::for(User::class)
    ->allowedSorts(['id' => 'users.id'])
    ->join('posts', 'posts.user_id', '=', 'users.id');

// GET /users?sort=id
// id sort will be aliased to the prefixed `users.id` in the query and thus not collide with the joined columns
```
Do you think this might solve your original problem?

PR #143, opened 2018-12-14
Escape models table from relation detection

dominikb:
This should fix the issue as described in [this comment](https://github.com/spatie/laravel-query-builder/pull/136#issuecomment-446900352).


freekmurze:
Thanks!

PR #154, opened 2019-01-15
Fix detection of false-positives for ignored values

### LINKED ISSUE #153: Boolean type filter ignored in 1.13.0
After updating to the recent version 1.13.0 I can no longer properly filter using boolean values. For example if I allow this filter:
`Filter::exact('is_visible'),`

This request:
`GET https://...api.../resource?filter[is_visible]=true` -> shows all visible resources

`GET https://...api.../resource?filter[is_visible]=false` -> ignores filter. Shows both visible and non visible

dominikb:
Fixes #153 

If values to be filtered for are `falsy`, they get ignored by `empty()`.

I added a test to detect this case and changed the check to an explicit `is_null()` check.

### DIFF
--- a/src/Filter.php
+++ b/src/Filter.php
@@ -38,7 +38,7 @@ class Filter
     {
         $valueToFilter = $this->resolveValueForFiltering($value);
 
-        if (empty($valueToFilter)) {
+        if (is_null($valueToFilter)) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


AlexVanderbist:
Awesome, thanks! Will tag this later today :D

Issue #130, opened 2018-11-20
Broken morph eager queries


rainerkent:
Hi, this solution seems to break the morph eager queries. As it was before it worked correctly for morph relations!

May be some like this could work:

```php
return [$fullRelationName => function ($query) use ($fields) {
    if(!($query instanceOf MorphTo)) {
        $query->select($this->prependFieldsWithTableName($fields, $query->getModel()->getTable()));
    }
}];
```

Or may be I got it all wrong! :(

_Originally posted by @rusowyler in https://github.com/spatie/laravel-query-builder/pull/118#issuecomment-436054718_

AlexVanderbist:
Feel free to PR a failing test for this one - if not I'll get back to this soon

rainerkent:
I'd also like to share my experience with this bug.

I have these three models:

``` php
class User 
{
    public function profile()
    {
        return $this->morphTo();
    }
}

class AdminProfile
{
    public function user()
    {
        return $this->morphOne(User::class, 'profile');
    }
}

class CustomerProfile
{
    public function user()
    {
        return $this->morphOne(User::class, 'profile');
    }
}
```

Controller:
``` php
return QueryBuilder::for(User::class)
    ->allowedIncludes('profile')
    ->get();
```

Now when I request:
`https://localhost:8000/api/users?include=profile`

I get QueryException:
`SQLSTATE[HY000]: General error: 1 no such table: users (SQL: select "users".* from "admin_profiles" where "admin_profiles"."id" in (1))"`

Including reverse (morphOne) relationship works as expected.

For now, I'm using rusowyler's suggestion as a workaround.

AlexVanderbist:
Thanks! Gonna merge the PR and provide a fix :)

AlexVanderbist:
Tagged in 1.13.1

PR #156, opened 2019-01-15
Failing test for including morphTo relationship

rainerkent:
Failing test showing that including morphTo relationships will throw QueryException caused by prepending selected fields with table name.


rainerkent:
Issue #130

Issue #163, opened 2019-01-27
Sorting Alias?


CDNRocket:
There is a way to use Alias also for sorting? Or this feature is available only for filters?

dominikb:
This is currently only available for filters.
@AlexVanderbist should alias be available on sorts for consistency?

freekmurze:
@dominikb Feel free to send a PR with tests for this.

PR #164, opened 2019-01-29
Enable the usage of an alias for a column name when using Sorts

dominikb:
This PR addresses #163

I'm not too fond of the solution [here](https://github.com/dominikb/laravel-query-builder/blob/777ac3c458211302df781ae0f2d8e9b218aff7ed/src/QueryBuilder.php#L157-L161)
```php
if (is_string($alias)) {
    return Sort::field(ltrim($alias, '-'), ltrim($sort, '-'));
}
return Sort::field(ltrim($sort, '-'));
```

Maybe someone can come up with a more readable solution?

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ $users = QueryBuilder::for(User::class)
 
 There may be occasions where it is not appropriate to expose the column name to the user.
 
-Similar to using [an alias when filtering]() you can do this with for sorts as well.
+Similar to using [an alias when filtering](#property-column-alias) you can do this with for sorts as well.
 
 The column name can be passed as optional parameter and defaults to the property string.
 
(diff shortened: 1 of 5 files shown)


AlexVanderbist:
Hey, thanks for another PR! Am I missing something or are there some changes to the `Sort` class missing?

AlexVanderbist:
I've added the missing code in `Sort`.

About that `is_string` clause that you don't like; I wouldn't mind just dropping the `allowedSorts(['sort' => 'alias'])` array notation. 
It's not supported for filters either. If you want to use any aliases you'll have to use the `Filter` and `Sort` classes which will be type hinted. This way there's no "magic" array syntax where you're unsure about whether the key or the value will be used as an alias, etc... 

Lmk what you think :)

dominikb:
Hey, thanks for the help there. I somehow forgot to commit half of my files :face_with_head_bandage: 

I like it better to use the actual `Sort` classes when aliasing a column. As you mentioned, it is more expressive in what it actually does and can be comprehended easier.

Thus I've decided to drop the array syntax and remove tests for it.

I've also updated the documentation to reflect theses adjustments.

PR #174, opened 2019-02-23
mark code as php

clarkeash:
I marked the code as php, and formatted the class code to be consistent with rest of the README


freekmurze:
Thanks!

Issue #176, opened 2019-02-27
Laravel 5.8 support


taai:
Yesterday Laravel 5.8 was released. I can't upgrade to newest Laravel version because this package supports Laravel versions up to 5.7.x. 😉 

By the way, does anyone know how to tell Composer to ignore version requirements of a specific package? I spent an hour and didn't find how to do it.

matthew-inamdar:
@taai I don't think you can ignore the package dependencies. you ignore the platform requirements (PHP version, extensions, etc.) with the `--ignore-platform-reqs` flag.

Your options are to:
1. Fork this repo and update the `composer.json` to support `laravel/framework 5.8` (be aware there may be breaking changes)
2. Wait until they release 5.8 support

AlexVanderbist:
We'll be tagging support for L5.8 in about half an hour :)

PR #177, opened 2019-02-27
Add Laravel 5.8 support

taai:
This PR addresses #176 
I'm pretty sure this works. Laravel 5.8 is not that different from 5.7. 😉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,18 +15,24 @@ matrix:
       env: LARAVEL='5.6.*' TESTBENCH='3.6.*'
     - php: 7.1
       env: LARAVEL='5.7.*' TESTBENCH='3.7.*'
+    - php: 7.1
+      env: LARAVEL='5.8.*' TESTBENCH='3.8.*'
     - php: 7.2
       env: LARAVEL='5.5.*' TESTBENCH='3.5.*'
     - php: 7.2
       env: LARAVEL='5.6.*' TESTBENCH='3.6.*'
     - php: 7.2
       env: LARAVEL='5.7.*' TESTBENCH='3.7.*'
+    - php: 7.2
+      env: LARAVEL='5.8.*' TESTBENCH='3.8.*'
     - php: 7.3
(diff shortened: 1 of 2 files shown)


AlexVanderbist:
Thanks!

telkins:
@AlexVanderbist Am I missing something...?  Seems like the tests fail because of the changes to the test code.

I was about to ask @taai if he needed help with this?  From the output, it looks like the test `setUp()` methods need to be updated to return `void`:
* Before: `protected function setUp()`
* After: `protected function setUp(): void`

It may not impact the actual code, but if the tests don't run....well, how will we know that things are OK?

I may have missed something, though....in which case, please let me know.  :-)

telkins:
I'm confused....travis-ci seems to be complaining about a line of code that appears correct in the master branch.  So, I must be missing something.

Enlightenment, anyone...?  ;-)

taai:
@telkins I'm pretty sure that the issue is with PHPUnit 8 which requires multiple methods like ```setUp()``` to have *return type declaration* ```: void```.
As I can see, @AlexVanderbist added PHPUnit 8 as a dev-dependency and later added also the type declarations in the unit tests. Now you have new problem – PHP 7.1 doesn't support *return type declarations* and that is why tests fail for PHP 7.1. Could this mean end of support for PHP 7.1 or is there *a way* to have and not to have *return type declarations* at the same time? 😐

telkins:
@taai Thx for the response.  Yes, I saw the issue with the missing return type declaration.  I guess I'm confused by why the PR was accepted with failing checks.  And, although I didn't dig into it, the master branch *has* the missing return type declaration....so I didn't know why it was missing from your PR nor did it didn't come from your PR.  😕 

FYI...return type declarations *are* supported by 7.1: http://php.net/manual/en/functions.returning-values.php.  The travis-ci error recognized the return type declaration but complained that subclass implementations of `setUp()` were missing the return type declaration.

Anyway...the process here seems to be what I'm struggling to understand.  I won't lose any sleep over it, though.  ;-)

Issue #178, opened 2019-02-27
Descending default sort no longer works


footballencarta:
With the release of 1.14.0 - the ability to set default sorts to descending no longer works.

For example:

**1.13.2**

```
QueryBuilder::for(User::class)
            ->defaultSort('-name')
            ->get();
```

Results in: 

`select * from "users" order by "name" desc, "name" desc`

**1.14.0**

```
QueryBuilder::for(User::class)
            ->defaultSort('-name')
            ->get();
```
Results in: 

`select * from "users" order by "-name" desc, "-name" desc`

The value being passed into default sorts no longer seems to parse the descending modifier correctly.

AlexVanderbist:
Thanks! The fix is merged and a version tagged 👍

duongtdvn:
I have upgraded to the latest version 1.15.2 and the bug still remain 🙁.

dominikb:
I'm unable to reproduce the original bug with not removing the leading dash (order by "-name" desc").

Are you actually referring to the order clause being added twice? This I can reproduce on 1.15.2 and I've already found the cause, but am still fiddling with a proper solution.

PR #180, opened 2019-02-27
Fix issue with descending default sort

axelitus:
When setting a descending default sort without specifying a column name, the column name was set to the property name with the dash prepended.

This pull request fixes that issue by setting the column name to the processed property name instead and adds a test for this use case.

### DIFF
--- a/src/Sort.php
+++ b/src/Sort.php
@@ -29,7 +29,7 @@ class Sort
 
         $this->defaultDirection = static::parsePropertyDirection($property);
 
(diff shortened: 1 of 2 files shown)


axelitus:
Fixes #178

AlexVanderbist:
Perfect! Thanks 👌

veneliniliev:
please merge this

AlexVanderbist:
Tagging this soon (30ish minutes) - you can use `dev-master` for now

Issue #169, opened 2019-02-05
Multiple default sorts?


matthew-inamdar:
Although you can provide multiple fields to sort by via the query string:

```
...?sort=first_name,last_name
```

It seems as if there is no way to provide multiple fields as the default sort. It only seems to support a single property:

```php
// ❌
QueryBuilder::for($baseQuery)
    ->defaultSort('first_name,last_name')
    ->allowedSorts('first_name', 'last_name');

// ❌
QueryBuilder::for($baseQuery)
    ->defaultSort('first_name', 'last_name')
    ->allowedSorts('first_name', 'last_name');

// ❌
QueryBuilder::for($baseQuery)
    ->defaultSort(['first_name', 'last_name'])
    ->allowedSorts('first_name', 'last_name');

// ✅
QueryBuilder::for($baseQuery)
    ->defaultSort('first_name')
    ->allowedSorts('first_name', 'last_name');
```

Can you confirm if this is the case? If so, I'm happy to go and open a PR to add this ability - I just want to check before I spend time doing so!

Cheers,
Matt

dominikb:
Hi Matt!

You're right on this one. The default sort on the `QueryBuilder` currently takes either a property name that gets converted into a `SortsField` object or an object implementing the `Sort` interface.

Currently, the only way to achieve this behaviour, is to create a specific custom Sort:

```php
class MultipleColumnSort implements Sort
{
    public function __invoke(Builder $query, $descending, string $properties) : Builder
    {
        foreach(str_split(',', $properties) as $property) {
            $query->orderBy($property, $descending ? 'desc' : 'asc');
        }
        
        return $query;
    }
}
```

Personally I think it would be great if default sorting would allow for the same syntax as adding new allowed sorts does.

matthew-inamdar:
Thanks @dominikb - I thought that was the case. I'll get a PR ready over the next few days to add this ability!

matthew-inamdar:
There's now a PR open to add this functionality.

PR #214, opened 2019-03-01
Multiple default sorts

matthew-inamdar:
This PR closes https://github.com/spatie/laravel-query-builder/issues/169

It allows the `defaultSort()` method to also accept an array of:

- Columns names as strings
- Custom sort classes

The method still allows for a single parameter to be passed for backwards compatibility.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,0 @@
-build
-composer.lock
-docs
-vendor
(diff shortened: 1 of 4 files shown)


AlexVanderbist:
Thanks!

Issue #179, opened 2019-02-27
Default sorts add sort twice when using toSql


footballencarta:
When using the toSql method - the defaultSort method adds the sort twice.

This is due to the toSql method calling back to the QueryBuilder getQuery method which seems to add the sorts again.

AlexVanderbist:
Thanks for reporting this! This was fixed in #219 and tagged in 1.17.3.

Issue #165, opened 2019-02-01
Custom sort


almas1992:
when I use custom sort like that 
```php 
QueryBuilder::for(User::class)
    ->allowedFilters('name')
    ->allowedSorts(Sort::custom('some',SomeSort::class))
    ->get()
```
the custom sort is not working , but if I use like 
```php
QueryBuilder::for(User::class)
    ->allowedSorts(Sort::custom('some',SomeSort::class))
    ->allowedFilters('name')
    ->get()
```
the custom sort is working.

Must I do `allowedSorts` before I do `allowedFilters`?

dominikb:
Hi almas1992!

Both methods can be used in any order you like and should not interfere with each other.

I tried to replicate the described behavior in a test but was unable to reproduce it.
Have you tried using a non-custom Sort? Maybe this has to do with the query your Custom-Sort is building?

dominikb:
I'm having a look at this.

I've already implemented a [failing test](https://github.com/dominikb/laravel-query-builder/commit/8e63873379f6f19770dd22c5ee432427d4a27318) for this scenario for everyone to play around with.

dominikb:
## The Problem
When adding new sorts to a `QueryBuilder` object, it checks if the new sort may be a duplicate and removes  those silently. This happens in `QueryBuilder::parseSorts()`.

When determining if the new Sort is a duplicate in `filterDuplicates($sorts)` the method `getQuery()` gets called. This method was overwritten in order to enable the possibility of default sorts and for allowing all sorts by default.
If the `QueryBuilder` gets created with a `Filter` at first, there are no sorts and `addDefaultSorts()` gets called.

Now, when checking for duplicate sorts, all sorts from the query are already added as `Sort::field` and later the actual custom sort will not be added to the list of allowed sorts.

```php
// Here we use the magic getters() to access the query property 'sort'
// These expressions are equal and then by default allow every sort passed in the query
// $this->request->sorts($this->defaultSort)
// $this->request->sorts
$this->allowedSorts = collect($this->request->sorts($this->defaultSort))
    ->map(function ($sort) {
    if ($sort instanceof Sort) {
        return $sort;
    }

    return Sort::field(ltrim($sort, '-'));
});
```


### Workaround
One way of avoiding this issue is to add a `Sort` as first chained method on the `QueryBuilder` instance.

### Solutions
[...]

PR #215, opened 2019-03-01
Remove duplicate parsing of sort clauses

dominikb:
Depending on the call order sorts could be parsed multiple times.

"Normal" sorts usually get only parsed once, but default sorts could be parsed twice (e.g. when calling `getQuery()` multiple times).

This resulted in strange queries as described in #179.

This PR attempts to fix the above mentioned issue and reduce duplicated code.

The implementation is not yet finished and will surely result in conflicts with the current PR #214.

I will keep on working on this and adapt it to the changes in PR #214 once/if it gets merged.

### DIFF
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -35,6 +35,9 @@ class QueryBuilder extends Builder
     /** @var \Illuminate\Http\Request */
     protected $request;
 
+    /** @var bool */
+    protected $sortsWereParsed = false;
+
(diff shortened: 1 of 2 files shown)


dominikb:
Fixes #217, #179, #165

VinceG:
@dominikb Any idea what the status on this?

dominikb:
I've opened another PR for the finished implementation. #219

But I dont know if any maintainer has had time to do a review.

PR #219, opened 2019-03-10
Remove duplicate parsing of sort clauses

### LINKED ISSUE #217: When defaultSort is set, any sort aliasing doesn't work
**Environment**

| Type | Version |
| ------------- | ------------- |
| Laravel  | 5.8.3 |
| PHP  | 7.2 |
| MySQL | 5.7 |
| Package version | 1.16.0 |

**Issue**
Having a `defaultSort()` set, when trying to add aliasing with `Sort::field()`, it throws an error : 
```sql
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'joined' in 'order clause' (SQL: select * from `users` order by `joined` asc)
```

**Steps to reproduce**
1. Install package, no configuration change
2. 
```php
   ->defaultSort('name')
   ->allowedSorts(Sort::field('joined', 'created_at'))
   ->get();
```
3. Request
`/users?sort=joined`

***Note***
It works without a default sort
```php
   ->allowedSorts(Sort::field('joined', 'created_at'))
   ->get();
```

dominikb:
Depending on the call order sorts could be parsed multiple times.

"Normal" sorts usually get only parsed once, but default sorts could be parsed twice (e.g. when calling getQuery() multiple times).

Fixes #217, #179, #165

_New PR because I'm not smart enough to change the state of my Draft PR_ :1st_place_medal:

### DIFF
--- a/src/Concerns/SortsQuery.php
+++ b/src/Concerns/SortsQuery.php
@@ -18,6 +18,14 @@ trait SortsQuery
     /** @var bool */
     protected $sortsWereParsed = false;
 
+    /**
+     * Per default, sorting is allowed on all columns if not specified otherwise.
+     * We keep track of those default sorts to purge them if, at a later point in time, allowed sorts are specified.
+     *
+     * @var array
+     */
+    protected $generatedDefaultSorts = [];
+
     public function allowedSorts($sorts): self
     {
         $sorts = is_array($sorts) ? $sorts : func_get_args();
@@ -76,15 +84,16 @@ trait SortsQuery
             return;
         }
 
+        $this->sortsWereParsed = true;
+
         if (! $this->allowedSorts instanceof Collection) {
             $this->addDefaultSorts();
+            $this->allowRepeatedParse();
+        } else {
+            $this->purgeGeneratedDefaultSorts();
         }
 
         $sorts = $this->request->sorts();
-
-        if ($sorts && ! $this->allowedSorts instanceof Collection) {
-            $this->addDefaultSorts();
-        }
 
         if ($sorts->isEmpty()) {
             optional($this->defaultSorts)->each(function (Sort $sort) {
@@ -102,8 +111,6 @@ trait SortsQuery
 
                 $sort->sort($this, $descending);
             });
-
-        $this->sortsWereParsed = true;
     }
 
     protected function findSort(string $property): ?Sort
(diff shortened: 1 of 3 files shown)


nickescobedo:
Yes, please. I noticed the other day too and I thought it was something I was doing.

VinceG:
@freekmurze @AlexVanderbist Hey Guys, can we please review this and merge?

AlexVanderbist:
Hey, thanks for the PR and sorry for leaving this open for so long. I've just pushed some changes that are causing a conflict for this PR but I'll fix those and merge this on Monday (or maybe this weekend).

dominikb:
I've merged the upstream changes back into my PR successfully, but one test is failing now.

The test __it_executes_the_same_query_regardless_of_the_order_of_applied_filters_or_sorts__ fails because, within the `__invoke()` method on the `FilterPartials` class, `getQuery()` gets called, which triggers the parsing of sorts and sets the flag to avoid repeated parsing of sorts.

I've yet to come up with a clean solution, but one solution could be to keep track of already parsed sort-clauses by name and filter them out the second time.
This, however, would still allow for the premature application of default sorts.

AlexVanderbist:
Hey, thanks again for all the good work. 

I've been thinking about changing the default behaviour of the fields and sorts in v2 of the package. Currently we can sort and select any fields by default, unless specified differently using `allowedFields` and `allowedSorts`. A side effect is the need to extend the `getQuery` method (and a couple others). In v2 all fields and sorts will have to be specified explicitly. That means we can update the query straight from the `allowedSorts` and `allowedFields` methods - and not from the `getQuery` method. 
If I understand correctly that should fix this problem as well... What do you think? 

If the above solution would work I don't mind having a somewhat hacky solution now and a very clean fix in v2.

dominikb:
It's a pleasure to work on this package :)

I'm all for the suggested change. Being explicit about which fields can be queried and sorted will help with straightening out some of the complexities introduced by the current flexibility. Such a change will surely fix the described issue.

I'll update the PR with the proposed changes soon and also add a disclaimer in the documentation to make users aware of this pitfall.

Do you have any more changes you'd like to see in v2 of this package?

AlexVanderbist:
Sounds good! I'll create an issue with some ideas for v2 and maybe move ideas from existing issues to that issue. I'll also set-up a branch for v2 as well.

Again, thanks for all the good work!

dominikb:
I've pushed my changes and all tests pass locally, but it seems that `json` methods in queries lead to errors in Travis CI.

AlexVanderbist:
Perfect, I'll take a look at the Travis tests before tagging a patch version with this fix 👍 Thanks again!

PR #222, opened 2019-03-19
Add support to get parameters from json of Illuminate\Http\Request

dexfs:
When I use 
```
$this->json('GET', '/endpoint', ['teste' => 'teste']) 
```
on my tests the parameters not getting by the QueryBuilderRequest correctly and It is because the Request put this ones in *json* property and QueryBuilderRequest get from query method.

### DIFF
--- a/src/QueryBuilderRequest.php
+++ b/src/QueryBuilderRequest.php
@@ -17,10 +17,10 @@ class QueryBuilderRequest extends Request
     {
         $parameter = config('query-builder.parameters.include');
 
-        $includeParts = $this->query($parameter);
+        $includeParts = $this->getPartsOfRequest($parameter);
 
         if (! is_array($includeParts)) {
-            $includeParts = explode(',', strtolower($this->query($parameter)));
+            $includeParts = explode(',', strtolower($this->getPartsOfRequest($parameter)));
         }
 
         return collect($includeParts)->filter();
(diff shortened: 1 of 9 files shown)


AlexVanderbist:
Hi! Thanks for the PR. Not checking the `json` in the request body is as intended so I'm not gonna merge this for now.

Laravel's `$this->json('GET', '/endpoint', ['data' => 123])` test helper will send a GET request with the data in the request **body**. However, GET requests should never have a request body (see https://tools.ietf.org/html/rfc2616#section-4.3). Even though you can, you actually shouldnt. In fact, I would advise using the `$this->getJson()`, `$this->postJson()`, etc... methods as they will more clearly indicate what parameters are expected/required.

Finally to fix your tests, you're probably gonna want to include any query builder parameters in the url like this:
```php
$this->getJson('/endpoint?filter[name]=dexfs')->assertSuccessful();
```

Good luck!

dexfs:
Thanks for your advice, You're right, the tests get better with correct method. 

Btw, I think the json send the data correct, not in its body, but the Request of Illuminate has a different behavior with this data and put it into json property. 

Oh, I'll follow your adivce and I'll change my tests.

Thank you.

dexfs:
@AlexVanderbist 
Just to note about the json method I mentioned in this PR.

https://github.com/laravel/framework/blob/5.8/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php#L306

The methods you advise to use they call the *json*  method internally.
The behavior  this PR "solve" is caused by the https://github.com/laravel/framework/blob/5.8/src/Illuminate/Http/Request.php#L400


Btw, thank you again.

PR #237, opened 2019-04-10
Allow sorting on related model fields

pulkitjalan:
It currently blocks sorting on a relationship because the regex doesn't allow for dot notation.

`->defaultSort('relation.field')`

### DIFF
--- a/src/ColumnNameSanitizer.php
+++ b/src/ColumnNameSanitizer.php
@@ -15,7 +15,7 @@ class ColumnNameSanitizer
      * Column names are alphanumeric strings that can contain
      * underscores (`_`) but can't start with a number.
      */
-    private const VALID_COLUMN_NAME_REGEX = '/^(?![0-9])[A-Za-z0-9_-]*$/';
+    private const VALID_COLUMN_NAME_REGEX = '/^(?![0-9])[.A-Za-z0-9_-]*$/';
 
     public static function sanitize(string $column): string
     {
--- a/src/ColumnNameSanitizer.php
+++ b/src/ColumnNameSanitizer.php
@@ -15,7 +15,7 @@ class ColumnNameSanitizer
      * Column names are alphanumeric strings that can contain
      * underscores (`_`) but can't start with a number.
      */
-    private const VALID_COLUMN_NAME_REGEX = '/^(?![0-9])[.A-Za-z0-9_-]*$/';
+    private const VALID_COLUMN_NAME_REGEX = '/^(?![0-9])[A-Za-z0-9._-]*$/';
 
     public static function sanitize(string $column): string
     {


pulkitjalan:
I'm struggling to do a test for this, some help with that would be great 😄

src/ColumnNameSanitizer.php
```diff
@@ -15,7 +15,7 @@ class ColumnNameSanitizer
      * Column names are alphanumeric strings that can contain
      * underscores (`_`) but can't start with a number.
      */
-    private const VALID_COLUMN_NAME_REGEX = '/^(?![0-9])[A-Za-z0-9_-]*$/';
+    private const VALID_COLUMN_NAME_REGEX = '/^(?![0-9])[A-Za-z0-9._-]*$/';
```
willishq: Does this dot need to be escaped in order to represent the string literal `.`?

pulkitjalan: It works as is

morloderex: it should be escaped sense currently it would allow any character

pulkitjalan: I'm happy to add the escape, but as far as I was aware this is fine. Have also tested it in practical use and it does work as intended and continues to block other characters.

Without escape
---
Screenshot 2019-04-11 at 04 17 25

With escape
---
Screenshot 2019-04-11 at 04 17 33

Lelectrolux: Inside a regex class (the square brackets), dots and most otherwise special characters lose their special meanings.

`.` Matches any char  
`[.]` only matches the dot  

You can check this [here](https://regex101.com/r/y0hl1u/1)

Lelectrolux:
@pulkitjalan 

I might be wrong, but I looked at the current tests, and it test the "regex doesn't match" case, mainly string with a " " (space) character. [Test here](https://github.com/spatie/laravel-query-builder/blob/master/tests/FieldsTest.php#L143). It's the only ref to the new `InvalidColumnName` exception I could find. 


On the other hand, there is no direct check that code matching the regex doesn't throw the new exception. It's implied in the other tests.


Unless we are told otherwise, I think it's ok to leave it like it is, without specific test for the "passes" case.


Side note, while reviewing the tests, I found [this test named it_can_fetch_sketchy_columns_if_they_are_allowed_fields](https://github.com/spatie/laravel-query-builder/blob/master/tests/FieldsTest.php#L62) which verify non regex matching fields are allowed if whitelisted.


This, I think, means there is a temporary solution (whitelist everything from other tables). But I can't run the code/check right now, I'll try later this week.

AlexVanderbist:
Thanks, looks good! I've added a test for this as well.

However, how are you planning on using this? You can't sort eloquent queries by related properties like that... Are you joining other tables on your query before sorting? What am I missing? 😅

pulkitjalan:
Yea we are joining a few tables to query against them and have custom filters for. Some of the sorting is done in the relationships.

Thanks for merging it 👍

Issue #235, opened 2019-04-09
Getting error on the latest update


abishekrsrikaanth:
@AlexVanderbist I just took and update of the latest version and I am getting an error on line https://github.com/spatie/laravel-query-builder/commit/3aa483b63c79d9fabcb8653fe837a7736eb93bea#diff-3591cf5c140b20d1bf5fa38e063bfdf7L84

Error `Method Spatie\QueryBuilder\QueryBuilderRequest::sort does not exist.`

Lydnasty:
https://github.com/spatie/laravel-query-builder/issues/224#issuecomment-475294329

abishekrsrikaanth:
@Lydnasty already have that. The issues seems to be the fact that the method `sort` was removed from the class `QueryBuilderRequest` but is being called on the line I linked on my initial post.

jayjfletcher:
I'm having the same issue, failing on the sort

AlexVanderbist:
A fix by @jayjfletcher was just merged and will be tagged soon

PR #238, opened 2019-04-11
[BugFix] QueryBuilderRequest

jayjfletcher:
The wip security fix merge introduced a bug with the removal of the sort method as mentioned in Issue #235 .

This should fix it.


AlexVanderbist:
Good catch! Looks like the tests were still passing because `$request->sort()` still exists as a (deprecated) macro... Tagging this soon.

Issue #258, opened 2019-04-29
Purging default sorts breaks the use of orderByRaw clauses


dominikb:
> The ability to sort NULLs either first or last. Right now, if I try to use an `orderByRaw` while using `QueryBuilder` I get the following exception:
> 
> ```json
> {
>     "message": "Undefined index: column",
>     "exception": "ErrorException",
>     "file": "vendor/spatie/laravel-query-builder/src/Concerns/SortsQuery.php",
>     "line": 165
> }
> ```
> (As mentioned in #244 by @mikemand)

This bug occurs due to the attempt to remove automatically generated sort clauses, that are prohibited once allowed sort clauses are specified.

Currently working on a fix, should be done until Wednesday.

If anybody else encounters this issue, please let me know how you specify your order clauses so I can test more thoroughly.

matthew-inamdar:
I've just realised I've created a duplicate issue in #266 - but leaving open as it contains a bit more info for a fix.

I'm thinking a simple check for if the order array to see if it contains a `type` key and if the value of that key is `Raw` then ignore it? (See my issue for background info)

Assuming this won't break anything, I'm happy to go ahead and create a PR at some point.

loonpwn:
Any word on this @dominikb?

Without digging too much into the code, could we simply check if the `column` index exists for some sort of hot fix?

AlexVanderbist:
Thanks for waiting. I've merged @dominikb's PR and will tag it as 1.17.4 in a couple of minutes.

PR #261, opened 2019-05-01
Check for availability of sort property before purging

dominikb:
Addresses issue #258


AlexVanderbist:
As usual, thanks @dominikb! 🥇

PR #265, opened 2019-05-11
execute query from params

emtudo:
If you have a command line, there is no request for which it is not possible to use QueryBuilder.

Then a query by parameters may be required

Example use:

```
use App\User;
use Spatie\QueryBuilder\QueryBuilder;

clsss UserRepository
{
    protected $allowedFilter = ['name'];

    public function search(array $params = [])
    {
        // $params => ['name' => 'test']
        return QueryBuilder::for(User::class, null, ['filter' => $params])
            ->allowedFilters($this->allowedFilter)
            ->get();

        // $params => ['filter' => ['name' => 'test']]
        // return QueryBuilder::for(User::class, null, $params)
            ->allowedFilters($this->allowedFilter)
            ->get();
    }
}
``

### DIFF
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -36,13 +36,18 @@ class QueryBuilder extends Builder
      *
      * @param string|\Illuminate\Database\Query\Builder $baseQuery Model class or base query builder
      * @param \Illuminate\Http\Request                  $request
+     * @param array                                     $params
      *
      * @return \Spatie\QueryBuilder\QueryBuilder
      */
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by AlexVanderbist** [comment]
Hi! Thanks for the PR. Have you tried creating a `Request` instance instead to pass in the parameters? A bit like the tests for this package:

```php
        $request = new Request([
            'filter' => ['name' => 'test'],
        ]);

        return QueryBuilder::for(TestModel::class, $request);
```

Issue #268, opened 2019-05-15
Sort::field with filter generate   error sql syntax


kadevland:
I have Product ,Category, and Price table
I have set filter sku name and customer filter to search on Price and Category table.
i define sort ,sku, name and sort alias update => update_at
When i sort on update or -update  without filter no problem 
when i sort on -update wiht filter i have error sql :
select .......  where ....... order by update desc ,update_at desc

dominikb:
Hi, thanks for opening this issue.

Can you please elaborate in a little more detail and maybe provide a code snippet to reproduce the issue?
I've tried to reproduce it myself, but as also suggested by [this test case](https://github.com/spatie/laravel-query-builder/blob/2144c04bc164068a45a2e5ddaf4b148c94f38b22/tests/QueryBuilderTest.php#L134) everything worked as intended.

Can you try calling `allowedSorts()` and `allowedFilters()` in different orders and see if the same error occurs?
What version of this package do you use?

ImadSalimi:
@dominikb but try it in the case where `name` is an alias for the column `username`, for example. Then you would get the SQL error about *name* not being a column

dominikb:
Thanks for pointing me in the right direction, I suppose I did not read carefully enough the first time.

I've tracked down the issue and submitted a PR.

Until it gets merged, a workaround can be to specify `allowedSorts()` first in the method-chain on the `QueryBuilder` instance.

AlexVanderbist:
Thanks for reporting this & thanks @dominikb for another great PR 👍 

Currently merging this and will be tagged as 1.17.4.

PR #272, opened 2019-05-22
Take leading dash into account when remembering generated sort columns

dominikb:
Addresses issue #268


ImadSalimi:
Thanks I really hope they merge it soon enough 😄 👍

PR #285, opened 2019-06-20
Eliminates empty value on partial match filter

denjaland:
Fixes issue where `/api/resource?filter[status]=1,2,3,` would return everything because of the trailing comma.  A partial match on empty value does not make sense and should be avoided.

It would be wise to of course eliminate the trailing comma when making the request, but just in case where it does get added, this fix eliminates unexpected results.


AlexVanderbist:
Thanks!

PR #299, opened 2019-06-25
Adds chunk method to query builder to parse sorts

marcoocram:
Since 1.17.3 version, when @AlexVanderbist fix duplication of sorts, if you use "chunk" method con query builder, the default sort doesn't affect because the chunk funcion use the enforceOrderBy function before get on queryBuilder and the parseSorts is not executed, so the default model sort is applied (id).

This PR adds the chunk function to ensure parse sorts before enforceOrderBy method is executed on chunk queries.

### DIFF
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -101,6 +101,13 @@ class QueryBuilder extends Builder
         return parent::simplePaginate($perPage, $columns, $pageName, $page);
     }
 
+    public function chunk($count, callable $callback)
+    {
+        $this->parseSorts();
+
+        return parent::chunk($count, $callback);
+    }
+
     /**
      * Add the model, scopes, eager loaded relationships, local macro's and onDelete callback
      * from the $builder to this query builder.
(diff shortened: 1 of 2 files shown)


freekmurze:
Could you add a test for this as well?

marcoocram:
yes! but i need some time ;)

marcoocram:
test added ^^

src/QueryBuilder.php
```diff
@@ -101,6 +101,13 @@ public function simplePaginate($perPage = null, $columns = ['*'], $pageName = 'p
         return parent::simplePaginate($perPage, $columns, $pageName, $page);
     }
 
+    public function chunk($count, callable $callback)
+    {
+        $this->parseSorts();
+
```
marcoocram: i've removed the `addAllRequestedFields` block because chunk calls to get method and aplly this requested fields on his function

AlexVanderbist:
Thanks!

PR #319, opened 2019-08-12
Align type-hint for for method with __construct.

ppx17:
Currently the `__construct` method type hints a `Illuminate\Database\Eloquent\Builder $builder`, but the static `for` method has a `\Illuminate\Database\Query\Builder` in its PHPDoc, that passes it to the `__construct()`. 

I've changed the Phpdoc of `for` to match the \Illuminate\Database\Eloquent\Builder used in `__construct()`. Now Phpstan seems satisfied and everything still works when I try it.


AlexVanderbist:
You're right, thanks!

PR #333, opened 2019-09-18
Updated pagination.md

ChanningDefoe:
By default the ->paginate() method does not pick up the parameters from Spatie Query builder. By using the appends method and grabbing the request query, you can add the parameters to the first_page_url, last_page_url, and next_page_url of the returned pagination JSON.


AlexVanderbist:
Thanks, great addition!

PR #337, opened 2019-09-22
Added method to ignore exception

emtudo:
Added method to ignore exception if only wanted to ignore invalid filters

### DIFF
--- a/src/Concerns/FiltersQuery.php
+++ b/src/Concerns/FiltersQuery.php
@@ -62,7 +62,7 @@ trait FiltersQuery
 
     protected function ensureAllFiltersExist()
     {
-        if (!$this->exceptionInvalidFilter) {
+        if (! $this->exceptionInvalidFilter) {
             return;
         }
 
--- a/src/Concerns/FiltersQuery.php
+++ b/src/Concerns/FiltersQuery.php
@@ -62,6 +62,10 @@ trait FiltersQuery
 
     protected function ensureAllFiltersExist()
     {
+        if (!$this->exceptionInvalidFilter) {
+            return;
+        }
+
         $filterNames = $this->request->filters()->keys();
 
         $allowedFilterNames = $this->allowedFilters->map(function (AllowedFilter $allowedFilter) {
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -12,6 +12,14 @@ use Spatie\QueryBuilder\Concerns\AppendsAttributesToResults;
 
 class QueryBuilder extends Builder
 {
+    /**
+     * Allow ignore invalid fields without throwing an exception
+     * if is false
+     *
+     * @var bool
+     */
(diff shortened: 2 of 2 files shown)


AlexVanderbist:
Thanks for the PR! What's the use-case for being able to pass additional, invalid filters?

emtudo:
Maintain semantics when passing filters, but treat differently before sending to QueryBuilder

gpibarra:
It would be good to also add in the configuration file "query-builder.php"

AlexVanderbist:
Thanks for the PR! I've manually merged this in, adding support to disable all types of invalid query exceptions (including sorts, filters, includes and appends). We'll tag a new version for this soon. Thanks!

AlexVanderbist:
Update on those change: I gave it a shot in 187043e4fe77d6abe2dda49332c02a5409f95eb0 and 48cde449d44aa33479978bbf271efbfa327d03c6 but no succes. If we just disable the exceptions we'll leave everyone's endpoints open for invalid queries (and thus SQL injection, weird stuff, etc). The correct solution would consist of not throwing the exception but also not parsing the invalid queries. Currently that's very hard to achieve without refactoring big parts of the package which is not something we're planning on. Open for PRs tho.

I know simply disabling the exceptions seems like a really easy solution, but I don't want to expose users of that package to security issues because they don't know the implications of disabling those exceptions.

If, for any reason, you need to allow filters that don't actually exist, I'd propose adding a custom `NullFilter` that does nothing to the query. You can then add your special filter in the allowedFilters array using the `NullFilter` and the exception will no longer be thrown. 

For example: 
```php
use Spatie\QueryBuilder\Filters\Filter;
use Illuminate\Database\Eloquent\Builder;

class NullFilter implements Filter
{
    public function __invoke(Builder $query, $value, string $property)
    {
        // do nothing
    }
}

// in your controller
QueryBuilder::for(User::class)->allowedFilters([
    'username', 
    'email',
    AllowedFilter::custom('special-filter', new NullFilter()),
])

// GET /users?filter[special-filter]=abc will no longer throw an exception as it's a valid filter.
```

ppisecky:
@AlexVanderbist 
What exactly are the security concerns here? From the brief look at the `FiltersQuery` trait it seems to me like disallowed filters are not processed as the `addFiltersToQuery` method iterates over the `allowedFilters` collection rather than `$request->filters()`. So only allowed filters should be even attempted to be processed as it stands, no?

But it does seem like the case is inverse for includes for example - requested includes are iterated over and paired with allowed includes. But here the include could just be skipped if disallowed, no?

Fields seem to work a little differently though. Is part of the problem a desire for consistency - i.e. not adding to feature to any query modifiers if it can't be in all of them?

slykar:
@AlexVanderbist It would be great if this could be documented. I need to filter multiple queries with different filters for the same request and I was looking for a way to not throw an exception, only apply the filters defined and silently ignore anything that was not allowed.

PR #342, opened 2019-09-25
Added support for nested array filtering

cappuc:
I've added two tests to check that filter's value can be a multidimensional array and changed the value filtering function.

`array_diff()` function fails on multidimensional array because it cast the value to string, so in that case I think that we should run diff on array keys instead of values with `Arr::except()`

### DIFF
--- a/src/AllowedFilter.php
+++ b/src/AllowedFilter.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\QueryBuilder;
 
+use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Spatie\QueryBuilder\Filters\Filter;
 use Spatie\QueryBuilder\Filters\FiltersExact;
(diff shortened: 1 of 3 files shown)


src/AllowedFilter.php
```diff
@@ -122,7 +123,11 @@ public function hasDefault(): bool
     protected function resolveValueForFiltering($value)
     {
         if (is_array($value)) {
-            $remainingProperties = array_diff($value, $this->ignored->toArray());
+            $isMultidimensional = count($value) !== count($value, COUNT_RECURSIVE);
```
cappuc: Maybe there is a better way to do this check

AlexVanderbist:
I like the idea a lot, thanks for the PR! Regarding the `array_diff` vs `Arr::except`: I think it's too confusing that the ignored values are sometimes keys and sometimes values. Maybe we can actually compare the contents of the multi-dimensional array with the contents in the `ignored` array? Might not be the most useful feature but at least it'll be consistent.

cappuc:
Yes, maybe it's better to keep consistency comparing the contents of multi-dimensional arrays but how can we allow to ignore all values inside a key? I think it can be a useful feature with multi-dimensional arrays.
Maybe using something like `['abc' => '*']`?

bberlijn:
It would be nice if multi dimensional arrays could be supported to make nested filtering possible.

We want to support `filter[key][operator]=value,value`, but this is not possible currently due to the `Notice: Array to string conversion` error.

PR #346, opened 2019-09-27
Split comma separated filter values on unspaced commas only (fixes #229)

### LINKED ISSUE #229: Filter on multiple string values containing commas
I have a big database with cars where I want to be able to do exact filtering on multiple model names. I do not have control over the formatting of these names. Some model names contains commas:
"911 RS 2, 2.7 lightweight", "911R", "550 Spyder"

Filtering on the multiple modal names by comma separation works as expected.
https://car-api/cars?filter[model]=911R%2C550%20Spyder -> "911R","550 Spyder"

Except for when the name contains a comma
https://car-api/cars?filter[model]=911%20RS%202%2C%202.7%20lightweight
 -> "911 RS 2, 2.7 lightweight"
It searches for two separate values.
`select * from cars where model in (?, ?) and....`

I know this is expected behavior, and it works as intended.

I tried escaping the non-value-separating commas with `\,` /  `%5C%2C`

I tried following https://github.com/spatie/laravel-query-builder/issues/139 but don’t want to lose the ability to do filtering on multiple variables. Is there a way to have individual variable containing commas, and still be filterable?

aphofstede:
Fixes most common cases of https://github.com/spatie/laravel-query-builder/issues/229
Commas used in values will usually have space around them so here we only treat commas without space around them as delimiters.

Old:
Filter value: `'Antwerpen, BE,Rotterdam, NL'`
Result: `['Antwerpen', ' BE', 'Rotterdam', ' NL']`

New:
Filter value: `'Antwerpen, BE,Rotterdam, NL'`
Result: `['Antwerpen, BE', 'Rotterdam, NL']`

Note: Using commas as delimiter hard-coded was left in-place, could be a new PR as suggested in linked issue.

### DIFF
--- a/src/QueryBuilderRequest.php
+++ b/src/QueryBuilderRequest.php
@@ -123,7 +123,7 @@ class QueryBuilderRequest extends Request
     private function explodeIgnoringWhitespace(string $delimiter, string $string): array
     {
         return preg_split(
-            '/' . preg_quote($delimiter) . '(?=\S)/',
+            '/'.preg_quote($delimiter).'(?=\S)/',
             trim($string, $delimiter),
             null,
             PREG_SPLIT_NO_EMPTY
--- a/src/QueryBuilderRequest.php
+++ b/src/QueryBuilderRequest.php
@@ -100,7 +100,7 @@ class QueryBuilderRequest extends Request
         }
 
         if (Str::contains($value, ',')) {
-            return explode(',', $value);
+            return $this->explodeIgnoringWhitespace(',', $value);
         }
 
         if ($value === 'true') {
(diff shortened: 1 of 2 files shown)


AlexVanderbist:
Thanks for the PR! Not gonna merge this for now as it's a bit too specific for this single use-case (also see #229)

aphofstede:
Thanks for evaluating it @AlexVanderbist , I understand. An alternative approach would be to add support for query strings like `?cities[]=Antwerpen, BE&cities[]=Rotterdam, NL`

AlexVanderbist:
@aphofstede sorry, I don't usually check replies in closed issues. I think you can already achieve this with a custom filter. E.g. `?filter[cities][]=Antwerpen, BE&filter[cities][]=Rotterdam, NL`. The custom filter will receive the cities as an array. You can then further process them based on the exploded string :)

PR #363, opened 2019-10-09
Update filter documentation.

denisrpriebe:
Custom filter implementations should return an instance of `\Illuminate\Database\Eloquent\Builder`.


AlexVanderbist:
Hi @denisrpriebe, this has actually recently been changed to not having to return the instance. I've updated this in your PR and will merge this now.

Thanks a lot!

PR #379, opened 2019-11-21
Fix #376: allow page size control

### LINKED ISSUE #376: Add `limit` parameter
Hey :)

This library is awesome, but one thing i've been missing is the ability to limit the amount of results. 

I think this would be pretty simple to add, so if you are happy I can PR :)

klimov-paul:
Fix #376 

Added separated methods for pagination, which allow page size be set by HTTP request parameter.

### DIFF
--- a/config/query-builder.php
+++ b/config/query-builder.php
@@ -3,7 +3,7 @@
 return [
 
     /*
-     * By default the package will use the `include`, `filter`, `sort`
+     * By default the package will use the `page`, `per-page`, `include`, `filter`, `sort`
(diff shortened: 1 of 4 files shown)


AlexVanderbist:
Hi, thanks for the PR! I'm gonna hold of merging this one in favour of the json-api-pagination package.  We'd rather not maintain the same feature in two different packages. 

Using the pagination package in combination with this one is documented here: https://docs.spatie.be/laravel-query-builder/v2/advanced-usage/pagination/

Thanks!

PR #383, opened 2019-11-21
Fix #382: add `FilterTrashed`

### LINKED ISSUE #382: Add `FilterTrashed`
It would be nice to support soft deleted records filtering out of the box.

It can be encapsulated into dedicated `FilterTrashed` filter, which will respond in way similar to the one Nova does. It would respond to `trashed` keyword: if it is empty - do not apply any scope, if it equals to 'with' - include trashed records, if equals to 'only' - display only trashed records.

klimov-paul:
Fix #382: add `FilterTrashed`


src/Filters/FiltersTrashed.php
```diff
@@ -0,0 +1,48 @@
...
+ * Usage example:
+ *
+ * ```php
+ * QueryBuilder::for(Item::class)
+ *     ->allowedFilters([
+ *         AllowedFilter::trashed(),
+ *         // ...
+ *     ]);
+ * ```
+ *
```
AlexVanderbist: Can we move these docs to the markdown files in the /docs directory? We haven't really got big doc blocks like this in the rest of the package.

AlexVanderbist:
LGTM! If someone can move the doc block to the docs directory I'll merge this in :)

PR #395, opened 2019-12-10
Handle empty or null includes in allowed includes array

georgebohnisch:
Allows the `allowedIncludes()`method to be a bit more flexible and improves on the Exception language if no includes exist at all for a query.


AlexVanderbist:
Thanks!

PR #398, opened 2019-12-25
Fix some AllowedSort docblocks

imliam:
These two docblocks were the wrong way around, each referencing the type the other should've.


freekmurze:
Thanks!

PR #403, opened 2020-01-06
Add ability to appends to relations.

crotanite:
(First pull request so I hope this is okay!)

Updated the AppendsAttributesToResults concern to add the ability to append to a relationship.


src/Concerns/AppendsAttributesToResults.php
```diff
@@ -23,11 +23,51 @@ public function allowedAppends($appends): self
...
-            return $result->append($this->request->appends()->toArray());
-        });
-    }
+	{
+	    $appends = $this->request->appends();
+	    return $results->each(function($item) use($appends)
+	    {
+	        $appends->each(function($append) use($item)
+	        {
+	            if(strpos($append, '.'))
```
AlexVanderbist: We like to use `Str::contains()` for added readability

src/Concerns/AppendsAttributesToResults.php
```diff
@@ -23,11 +23,51 @@ public function allowedAppends($appends): self
...
+	            {
+	                $item->append($append);
+	            }
+	        });
+	    });
+	}
+
+	private function appendLoop($item, $relation, $subs)
+	{
+		if($item->$relation !== null)
```
AlexVanderbist: I think this will cause the relationship to be loaded, even though it wasn't requested (or allowed). Can we use `$item->relationLoaded($relation)` instead?

src/Concerns/AppendsAttributesToResults.php
AlexVanderbist: Nitpick: we like to use `Str::contains()` for added readability

src/Concerns/AppendsAttributesToResults.php
```diff
@@ -23,11 +23,51 @@ public function allowedAppends($appends): self
...
+		        	$item = $this->appendLoop($item, $relation, $subs);
+	            } else
+	            {
+	                $item->append($append);
+	            }
+	        });
+	    });
+	}
+
+	private function appendLoop($item, $relation, $subs)
```
AlexVanderbist: Can we use typehints here? `$item, string $relation, Collection $subs`

src/Concerns/AppendsAttributesToResults.php
```diff
@@ -23,11 +23,51 @@ public function allowedAppends($appends): self
...
-    }
+	{
+	    $appends = $this->request->appends();
+	    return $results->each(function($item) use($appends)
+	    {
+	        $appends->each(function($append) use($item)
+	        {
+	            if(strpos($append, '.'))
+	            {
+		        	$subs = collect(explode('.', $append));
```
AlexVanderbist: Can we rename `$subs` to `$subAppends` or `$nestedAppends` to make it easier to undertand.

src/Concerns/AppendsAttributesToResults.php
```diff
@@ -23,11 +23,51 @@ public function allowedAppends($appends): self
...
+		        	$relation = $subs->shift();
+
+		        	$item = $this->appendLoop($item, $relation, $subs);
```
AlexVanderbist: We don't need to store the returned value in `$item`
```diff
-		        	$item = $this->appendLoop($item, $relation, $subs);
+		        	$this->appendLoop($item, $relation, $subs);
```

src/Concerns/AppendsAttributesToResults.php
```diff
@@ -23,11 +23,51 @@ public function allowedAppends($appends): self
...
+		        {
+		            $item->$relation->append($sub);
+	       		}
+	    	} else
+	    	{
+	    		$sub = $subs->shift();
+	    		$item->$relation = $this->appendLoop($item->$relation, $sub, $subs);
+	    	}
+	    }
+	    return $item;
```
AlexVanderbist: We don't need to return `$item`. It's not used anywhere else :)

chrisidakwo:
@AlexVanderbist Please what's the current status of this?

Sharptsa:
@chrisidakwo I think it is still missing tests.

AlexVanderbist:
PR is indeed still missing tests

crotanite:
Might try to tackle tests tomorrow. I think it only needs one(?): it_can_append_to_relation.

Any others you'd suggest?
@AlexVanderbist

freekmurze:
Closing due to inactivity

Issue #404, opened 2020-01-06
setArrayValueDelimiter not working?


francoism90:
Thanks for merging #369. After some testing, I don't think it is working at the moment.

https://docs.spatie.be/laravel-query-builder/v2/features/filtering/ - Mentions `setArrayDelimiter`, but shouldn't it be `setArrayValueDelimiter`? 

Anyway, I've created the following middleware:

```
<?php

namespace App\Http\Middleware;

use Closure;
use Spatie\QueryBuilder\QueryBuilderRequest;

class SemicolonDelimiterMiddleware
{
    /**
     * Handle an incoming request.
     *
     * @param  \Illuminate\Http\Request  $request
     * @param  \Closure  $next
     * @return mixed
     */
    public function handle($request, Closure $next)
    {
        QueryBuilderRequest::setArrayValueDelimiter(';');

        return $next($request);
    }
}
```

Results in:

```
Requested include(s) `model,tags` are not allowed. Allowed include(s) are `model, modelCount, tags, tagsCount`.
```

MediaController.php

```

class MediaController extends Controller
{
    /**
     * @return MediaResource
     */
    public function index()
    {
        $query = QueryBuilder::for(Media::class)
            ->allowedIncludes(['model', 'tags'])
            ....
            ->jsonPaginate();

        return MediaResource::collection($query);
    }
...
```

It doesn't seem to matter if using a middleware or directly using a controller.

Thanks, if more info is needed, please let me know. :)

freekmurze:
@dominikb could you have a look at this?

francoism90:
@freekmurze @dominikb I just had a thought and think I solved my own (maybe stupid) question:
```
const response = await Vue.axios.get('media', {
      params: {
        include: 'model,tags' // <-- this needs to be model;tags as well
      }
    })

return response.data
```

It doesn't seem to only affect filters, but also includes, sorters (if possible), etc. as well.

dominikb:
@francoism90 Your guess is right, this is indeed how it works. 
https://github.com/spatie/laravel-query-builder/blob/8d4949f8539f363ff91633ce655462295aaca246/src/QueryBuilderRequest.php#L28-L41

Maybe I should have mentioned that in the documentation. I will send a pull request that clarifies this!

francoism90:
@dominikb This seems to be merged, thanks for the clarification. :)

PR #406, opened 2020-01-07
Move and improve documentation for array value delimiter

dominikb:
Make it clear, that this delimiter applies not only to filters but also to includes or sorts.

Inspired by #404


freekmurze:
Thanks!

Issue #412, opened 2020-01-13
[Bug] AllowedInclude::relationship and AllowedInclude::count can't be used


Sharptsa:
It is currently impossible to use in `allowedIncludes()` : 

- `AllowedInclude::relationship('name', 'internalName')` for aliasing relationships 
- `AllowedInclude::count('nameCount', 'internalNameCount')` as [stated in the docs](https://docs.spatie.be/laravel-query-builder/v2/features/including-relationships/#including-related-model-count)

While aliasing sort with `AllowedSort::field('fieldAlias', 'field')` works fine.

Example:
```php
$models = QueryBuilder::for(Model::class)
    ->allowedIncludes([
        AllowedInclude::count('relationshipCount'),
        AllowedInclude::relationship('relationshipAlias', 'relationship')
    ])
    ->get();
```
Trying to use those includes will result in the following error:
```json
{
    "message": "Requested include(s) `relationshipCount, relationshipAlias` are not allowed. Allowed include(s) are `[{}], [{}]Count, [{},{}], [{},{}]Count`.",
    "statusCode": 400
}
```

I believe the issue is located here:

https://github.com/spatie/laravel-query-builder/blob/d0d82095138f961241c81efe0a40147009ad6d4d/src/Concerns/AddsIncludesToQuery.php#L22-L24

Because `AllowedInclude::relationship()` and `AllowedInclude::count('nameCount', 'internalNameCount')` already return a Collection, it never passes in this condition

dominikb:
Here's a failing test to reproduce this issue:

```php
/** @test */
public function it_can_alias_multiple_allowed_includes()
{
    $request = new Request([
        'include' => 'relatedModelsCount,relationShipAlias'
    ]);

    $models = QueryBuilder::for(TestModel::class, $request)
        ->allowedIncludes([
            AllowedInclude::count('relatedModelsCount'),
            AllowedInclude::relationship('relationShipAlias', 'otherRelatedModels')
        ]);

    dd($models->first());
}
```

```
Spatie\QueryBuilder\Exceptions\InvalidIncludeQuery : Requested include(s) `relatedModelsCount, relationShipAlias` are not allowed. Allowed include(s) are `[{}], [{}]Count, [{},{}], [{},{}]Count`.
 /Users/dominik/code/laravel-query-builder/src/Exceptions/InvalidIncludeQuery.php:30
 /Users/dominik/code/laravel-query-builder/src/Concerns/AddsIncludesToQuery.php:71
[...]

PR #414, opened 2020-01-14
Fix alias for allowed includes

dominikb:
Addresses issue #412

### DIFF
--- a/src/AllowedInclude.php
+++ b/src/AllowedInclude.php
@@ -31,13 +31,24 @@ class AllowedInclude
         $internalName = Str::camel($internalName ?? $name);
 
         return IncludedRelationship::getIndividualRelationshipPathsFromInclude($internalName)
-            ->flatMap(function (string $relationship) use ($name, $internalName): Collection {
-                return collect([
-                    new self($relationship, new IncludedRelationship(), $relationship === $internalName ? $internalName : null),
-                ])
-                    ->when(! Str::contains($relationship, '.'), function (Collection $includes) use ($internalName, $relationship) {
-                        return $includes->merge(self::count("{$relationship}Count", $relationship === $internalName ? "{$internalName}Count" : null));
-                    });
+            ->zip(IncludedRelationship::getIndividualRelationshipPathsFromInclude($name))
+            ->flatMap(function ($args): Collection {
+                [$relationship, $alias] = $args;
+
+                $includes = collect([
+                    new self($alias, new IncludedRelationship, $relationship),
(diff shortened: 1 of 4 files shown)


Sharptsa:
@dominikb Thanks for taking the time !

I tried this on my codebase, works great with simple relationship and counts but fails with nested relationship.

```php
AllowedInclude::relationship( 'relationshipAlias', 'relatedModel.nestedRelatedModel');
```
Produces this error : 
```
Argument 1 passed to Spatie\\QueryBuilder\\AllowedInclude::__construct() must be of the type string, null given, called in /var/www/back/vendor/spatie/laravel-query-builder/src/AllowedInclude.php on line 38 at /var/www/back/vendor/spatie/laravel-query-builder/src/AllowedInclude.php:22)"
```

Though, now that I'm thinking about it, I don't know what is the intended behavior aliasing a nested relationship.
And what expected includes it should produce.

- `relatedModel` and `relationshipAlias` ?
- just `relationshipAlias` and ignore the traversing relationship ?
- something else ?

dominikb:
Hey, thanks for testing my PR.

I, maybe wrongfully, assumed that an aliased relationship would also include the previous model name:

```php
// This should work
AllowedInclude::relationship('firstModelAlias.secondModelAlias', 'firstModel.secondModel')

// but this would not
AllowedInclude::relationship('secondModelAlias', 'firstModel.secondModel')
```
The logic could be extended to handle both cases, by determining if the alias has a __dot__ in its name.

I'll update the PR shortly. The maintainers then need to make a decision which behavior we want to support.

Sharptsa:
Ah ok, I tested
```php
AllowedInclude::relationship('firstModelAlias.secondModelAlias', 'firstModel.secondModel')
```
And it works fine indeed. :+1: 

I was hoping this fix would also allow me to alias deep nested relationships to simplify the includes I have to pass in my requests.
But I guess this is a bit out of the scope of the issue I reported. This would be more suited as a feature/enhancement than a bugfix.

Would be great to see what the maintainers think.

Maybe for now we should just throw a proper exception if the depth of the alias and nested relationship don't match.

AlexVanderbist:
Thanks you both for looking into this. I'm gonna merge this first, just to fix the issue at hand. 

Afterwards, feel free to open a new issue to discuss aliasing nested relationships. As a start to that discussion: personally I've never felt the need for hiding nested relationships as the response is always gonna reflect that nested structure. I can see how it's useful for more complex API's too tho.

Issue #146, opened 2019-01-02
Add support for Eloquent models in scope using filters


thijsw:
It would be nice when models are automatically retrieved if they are type-hinted in the scope function.

For a URI like **/filter?[user]=1**, the package would retrieve user with ID = 1 and use the model in calling the scope method. When the model is not found, a 404 can be generated.

In Eloquent model:
```php
public function scopeUser(Builder $query, User $user): Builder
{
    return $query->has(/* ... */);
}
```

instead of:
```php
public function scopeUser(Builder $query, $id): Builder
{
    return $query->has(/* ... */);
}
```

dominikb:
I tried to implement this and it was possible without lots of overhead. [See Gist](https://gist.github.com/dominikb/a4340cf33d35bace9e77def9d0e97ee0)

However, I don't know if this would be a case better handled by using a custom filter?

AlexVanderbist:
I think one of my colleagues (@rubenvanassche) implemented this in a client project as well. I'm not sure what his final thoughts were but I'll discuss this with him tomorrow. 

Either way, I wouldn't mind having this in a new, custom filter.

AlexVanderbist:
Only a couple months of delay 🙈 

If someone want to add this to the `ScopeFilter` in a PR with documentation, I'll be happy to merge it.

@dominikb's comment above contains a good starting point. Here's another implementation from a client project we did: https://gist.github.com/AlexVanderbist/043a6065e43eda616148315ef02583c7

DaniloBenevides:
Is anyone working on this issue?

michapietsch:
@thijsw I'd rather not generate a 404. IMO it should return an empty result.

@DaniloBenevides I'd take it on. 😊

michapietsch:
Sorry, just realized the PR by @50bhan.

If he doesn't follow up on this, I'd like to offer my help.

Also, I think there is an issue with primary keys other than ```id``` and the implementation ```is_numeric($value)```. The type hinting is already done and the scope gets an Eloquent model instance, but if we work with the given identifier, we have to accept all options, right?

As mentioned above, I'd also not ```findOrFail```. Anyway, as I see it, a test for this case is missing. But what if we were to allow multiple options like ```filter[user]=1,3```, and user 1 didn't exist, but user 3?

But first I'll wait a bit and leave it to @50bhan.

AlexVanderbist:
Some great ideas, @michapietsch. Let's continue this discussion in the PR :)

PR #415, opened 2020-01-17
Support of type hinted scope parameters

50bhan:
Add ability of resolving type hinted scope parameters (issue #146)

### DIFF
--- a/src/Filters/FiltersScope.php
+++ b/src/Filters/FiltersScope.php
@@ -2,12 +2,12 @@
 
 namespace Spatie\QueryBuilder\Filters;
 
-use Spatie\QueryBuilder\Exceptions\InvalidFilterValue;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use ReflectionObject;
+use Spatie\QueryBuilder\Exceptions\InvalidFilterValue;
 
 class FiltersScope implements Filter
 {
@@ -32,11 +32,12 @@ class FiltersScope implements Filter
                 continue;
             }
 
-            $model = $parameter->getClass()->newInstance();
-            $index = $parameter->getPosition() - 1;
-            $value = $values[$index];
+            $model   = $parameter->getClass()->newInstance();
+            $keyType = $model->getKeyType();
+            $index   = $parameter->getPosition() - 1;
+            $value   = $values[$index];
 
-            if (is_numeric($value)) {
+            if ($keyType === 'string' || ($keyType === 'int' && is_numeric($value))) {
                 $result = $model::find($value);
 
                 if ($result === null) {
--- a/src/Filters/FiltersScope.php
+++ b/src/Filters/FiltersScope.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\QueryBuilder\Filters;
 
+use Spatie\QueryBuilder\Exceptions\InvalidFilterValue;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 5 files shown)


michapietsch:
@50bhan Are you going to follow up on this? Else I'd like to offer to help.

50bhan:
@michapietsch Thanks for the offer. Actually I think this feature is done and ready to merge. But doing what is up to @freekmurze and @AlexVanderbist. Maybe there is a policy for adding new features! I would appreciate if they enlighten us.

AlexVanderbist:
Hey, my bad fir not following this up sooner. I'll look into this PR when im at the office later today. Thanks!

michapietsch:
@50bhan Thanks for your work and sorry for rushing forward

src/Filters/FiltersScope.php
```diff
@@ -3,17 +3,43 @@
...
         $scope = Str::camel($property);
 
         $values = Arr::wrap($values);
+        $values = $this->resolveTypeHintedParameters($query, $values, $scope);
 
-        $query->$scope(...array_values($values));
+        return $query->$scope(...$values);
+    }
+
+    protected function resolveTypeHintedParameters(Builder $query, $values, string $scope): array
```
AlexVanderbist: Can we rename this to `resolveParameters`? 90% of the time they're just gonna be regular params instead of type hinted parameters

AlexVanderbist: Same for the `$scopeParameters` variable :)

src/Filters/FiltersScope.php
```diff
@@ -3,17 +3,43 @@
...
+            if (! optional($scopeParameter->getClass())->isSubclassOf(Model::class)) {
+                continue;
+            }
+
+            $model = $scopeParameter->getClass()->newInstance();
+            $index = $scopeParameter->getPosition() - 1;
+            $value = $values[$index];
+
+            if (is_numeric($value)) {
+                $values[$index] = $model::findOrFail($value);
```
AlexVanderbist: `findOrFail` will return a 404 response. This is not ideal for a query builder endpoint. I'm not sure what the proper response would be.

@michapietsch mentioned an empty result set but that might be confusing as to why it's empty. Maybe we can just throw a custom exception that explains the error?

50bhan: @AlexVanderbist I agree with custom exception. Maybe with 422 status code?

AlexVanderbist: Sounds like a plan!

src/Filters/FiltersScope.php
```diff
@@ -3,17 +3,43 @@
...
+        foreach ($parameters as $parameter) {
+            if (! optional($parameter->getClass())->isSubclassOf(Model::class)) {
+                continue;
+            }
+
+            $model = $parameter->getClass()->newInstance();
+            $index = $parameter->getPosition() - 1;
+            $value = $values[$index];
+
+            if (is_numeric($value)) {
```
dominikb: Would it make sense for this line to take into account the `$keyType` property on the model instance?

Models using a string primary key (e.g UUID) can not be resolved currently.

Something like:
```php
$keyType = $model->getKeyType();

if ($keyType == 'string' || ($keyType == 'int' && is_numeric($value))
{
}
```

50bhan: @dominikb Yes, of course! Thank you for the tip :)

50bhan:
@AlexVanderbist All changes has been done. Please check it out and let me know if something else is needed.

AlexVanderbist:
@50bhan thanks for the changes, everything looks great. 

One final thought regarding the (non-)numeric keys: would it make more sense to use `$model->resolveRouteBinding()`? This way the parameter resolving works the same way as their model resolving for routes. Any thoughts? (@dominikb feel free to chip in as well :) )

50bhan:
@AlexVanderbist Thanks. It's way better now.

PR #433, opened 2020-02-13
Update including-relationships.md

Naoray:
Update Eloquent api resources and conditional relationships url to point to latest documentation version


freekmurze:
Thanks!

PR #435, opened 2020-02-25
Fix for Lumen

rudashi:
Lumen doesn't have `config_path` function, so when you use console for migration or seed it crash.


freekmurze:
Thanks!

PR #454, opened 2020-03-21
check non existing relationships

pkboom:
This PR enables to check if relationships exist for includes. I will add a test and a exception if this PR is considered ok. Not sure checking for relationships at this level would be enough. Welcome any comments. Thanks. I've read source code of many packages you guys created and it's helped me a lot. I appreciate it.

### DIFF
--- a/src/Concerns/AddsIncludesToQuery.php
+++ b/src/Concerns/AddsIncludesToQuery.php
@@ -78,6 +78,16 @@ trait AddsIncludesToQuery
             throw InvalidIncludeQuery::includesNotAllowed($diff, $allowedIncludeNames);
         }
 
-        // TODO: Check for non-existing relationships?
+        $methods = get_class_methods($this->getModel());
+
+        $methods = collect($methods)->reject(function ($method) {
+            return method_exists('Illuminate\Database\Eloquent\Model', $method);
+        });
+
+        $diff = $includes->diff($methods);
+
+        if ($diff->count()) {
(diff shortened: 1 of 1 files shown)


src/Concerns/AddsIncludesToQuery.php
```diff
@@ -78,6 +78,16 @@ protected function ensureAllIncludesExist()
             throw InvalidIncludeQuery::includesNotAllowed($diff, $allowedIncludeNames);
         }
 
-        // TODO: Check for non-existing relationships?
+        $methods = get_class_methods($this->getModel());
+
+        $methods = collect($methods)->reject(function ($method) {
```
AlexVanderbist, line 83: `$methods` can be inlined in `collect()` here to avoid confusion over declaring the same variable twice :)

src/Concerns/AddsIncludesToQuery.php
```diff
@@ -78,6 +78,16 @@ protected function ensureAllIncludesExist()
...
+            return method_exists('Illuminate\Database\Eloquent\Model', $method);
```
AlexVanderbist, line 84: Please use `Model::class` here instead of the string representation

src/Concerns/AddsIncludesToQuery.php
AlexVanderbist, line 83: `$availableRelationships` might be a better name for the `$methods` variable

pkboom:
Opps. Files have been deleted. I will make a PR again.

PR #463, opened 2020-04-11
Add appends to paginator instances

malinushj:
HI!
I was testing out this branch because I really needed the **BelongsToMany** support from #443 but using it on a paginator I noticed it was missing the appends.
This PR will fix that (quite an easy fix). 🤞


AlexVanderbist:
Good PR, thanks a lot!

PR #465, opened 2020-04-14
Check non existing relationships

pkboom:
This PR enables to check if relationships exist for includes.

### DIFF
--- a/src/AllowedInclude.php
+++ b/src/AllowedInclude.php
@@ -69,6 +69,11 @@ class AllowedInclude
         return $this->name;
     }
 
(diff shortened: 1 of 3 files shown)


src/AllowedInclude.php
```diff
@@ -69,6 +69,11 @@ public function getName(): string
         return $this->name;
     }
 
+    public function getInterlName(): string
```
malinushj, line 72: @pkboom Typo.. `public function getInternallName(): string`

pkboom: Thanks.

Issue #466, opened 2020-04-17
Integrity constraint violation: 1052 Column &#039;column&#039; in where clause is ambiguous


KillWil:
Hello,

Filtering model field and in the same time with scope and join may cause ambigous in where clause.

**How to reproduce :**
1 table 'table_one' with a code field
1 another table 'table_two' with a code field
1 model 'table_one' with a scope 'MyScope' (I'm not using whereHas functions which is too long to execute)

```
public function scopeMyScope($query) {
         return $query->join("table_two", 'table_one.code', '=', 'table_two.code');
}
```

Then call a get request :
`GET myroute?filter[code]=15&filter[my_scope]`

**Solution :**
Fully qualified with the table name the properties fields in Exact Filter file :
/laravel-query-builder/src/Filters/FiltersExact.php

**$query->getModel()->getTable().'.'.**

```
public function __invoke(Builder $query, $value, string $property)
    {

        if ($this->addRelationConstraint) {
            if ($this->isRelationProperty($query, $property)) {
                $this->withRelationConstraint($query, $value, $property);

                return;
            }
        }

        if (is_array($value)) {
            $query->whereIn($query->getModel()->getTable().'.'.$property, $value);

            return;
        }

        $query->where($query->getModel()->getTable().'.'.$property, '=', $value);
    }

```
I can make a pull request if you want.

Thanks a lot,

AlexVanderbist:
Thanks! Let's continue this issue in your PR.

PR #467, opened 2020-04-17
Ambiguous filters exact

KillWil:
Refers to [issue 466](https://github.com/spatie/laravel-query-builder/issues/466)

### DIFF
--- a/src/Filters/FiltersExact.php
+++ b/src/Filters/FiltersExact.php
@@ -29,7 +29,7 @@ class FiltersExact implements Filter
         }
 
         if (is_array($value)) {
-            $query->whereIn($query->getModel()->getTable().'.'.$property, $value);
+    $query->whereIn($query->qualifyColumn($property), $value);
 
             return;
         }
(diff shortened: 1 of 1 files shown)


lexdewilligen:
Please merge this! Really need it

nickpoulos:
Just checking in on the status of this PR? We are also experiencing the same issue.

src/Filters/FiltersExact.php
```diff
@@ -29,12 +29,12 @@ public function __invoke(Builder $query, $value, string $property)
         }
 
         if (is_array($value)) {
-            $query->whereIn($property, $value);
+            $query->whereIn($query->getModel()->getTable().'.'.$property, $value);
```
BrandonSurowiec, line 32:
```suggestion
    $query->whereIn($query->qualifyColumn($property), $value);
```

AlexVanderbist:
Hi, thank you for reminding me. Tests for this PR do seem to be failing, I'll look into this right now.

Personally I'm still on the fence about this. I understand that combining joins with filters causes issues but this is also true for basic Eloquent. This package isn't meant to solve these kind of issues, it's just meant as a way to access Eloquent from the URL.
For example, the problem described in the issue (#466) could be fixed by specifying the table in the underlying `where` like this: `Model::query()->join('other_models', 'other_models.code', 'models.code)->where('other_models.code', 'foo')`. This directly applies to this package as well: `GET /models?filters[other_models.code]=foo`.

PR #469, opened 2020-04-19
Fixes issue resolving filter scope parameters

### LINKED ISSUE #432: SoftDelete scopes cause ReflectionException
When using version 2.6.* to 2.7.0 I am no longer able to add the soft delete scopes as an allowed filter.

```
QueryBuilder::for(User::class))
    ->allowedFilters([AllowedFilter::scope('only_trashed')])
    ->get()
```

This results in a "Method scopeOnlyTrashed does not exist" ReflectionException in
https://github.com/spatie/laravel-query-builder/blob/366aaca9cf6195357812f01f412fe878cec414f5/src/Filters/FiltersScope.php#L26-L28

My guess is, this is caused by the magic inside of the [SoftDeletingScope](https://github.com/illuminate/database/blob/master/Eloquent/SoftDeletingScope.php) that isn't allowing reflection to see the scopes added by the SoftDeletes trait.

I reverted my version back to 2.5.1 to temporarily resolve the issue.

ampersand:
Is this a viable solution to fix #432 ?

### DIFF
--- a/src/Filters/FiltersScope.php
+++ b/src/Filters/FiltersScope.php
@@ -6,6 +6,7 @@ use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
+use ReflectionException;
 use ReflectionObject;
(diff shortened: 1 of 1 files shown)


AlexVanderbist:
Looks good! Thanks for the PR :)

PR #497, opened 2020-07-22
Fix issue with inability to instantiate QueryBuilder with Relation object

golonix:
Currently, it is impossible to instantiate `QueryBuilder` object with `Relation`-like object/query. This is due typehint in `Illuminate\Database\Eloquent\Builder::__construct` method which is called by `QueryBuilder::__construct()` method. Eloquent builder object expects `\Illuminate\Database\Query\Builder` object as `$query` param.

```
    public function __construct($builder, ?Request $request = null)
    {
        parent::__construct(clone $builder->getQuery());

        $this->initializeFromBuilder($builder);

        ...
    }
```
When `$builder` is of `\Illuminate\Database\Eloquent\Relations\Relation`, `$builder->getQuery()` will return `\Illuminate\Database\Eloquent\Builder` instead of requested `\Illuminate\Database\Query\Builder`.

This MR fixes that.

### DIFF
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -29,6 +29,10 @@ class QueryBuilder extends Builder
      */
     public function __construct($builder, ?Request $request = null)
     {
+        if ($builder instanceof \Illuminate\Database\Eloquent\Relations\Relation) {
+            $builder = $builder->getQuery();
+        }
+
         parent::__construct(clone $builder->getQuery());
 
         $this->initializeFromBuilder($builder);
--- a/tests/QueryBuilderTest.php
+++ b/tests/QueryBuilderTest.php
@@ -172,31 +172,20 @@ class QueryBuilderTest extends TestCase
     /** @test */
     public function it_can_get_a_query_from_a_relationship()
     {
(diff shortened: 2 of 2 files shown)


src/QueryBuilder.php
```diff
@@ -29,6 +29,10 @@ class QueryBuilder extends Builder
      */
     public function __construct($builder, ?Request $request = null)
     {
+        if ($builder instanceof \Illuminate\Database\Eloquent\Relations\Relation) {
```
dominikb, line 32: You could add a comment to explain why this conditional exists. Without the context of this PR the repeated call to `$builder->getQuery()` could appear a little bit confusing.

AlexVanderbist:
Thanks for the PR! Sorry this took so long but I believe this has since been fixed in version 3 of the package. Feel free to open a new issue if you're still experiencing issues with this. Thanks!

golonix:
@AlexVanderbist ah, no worries. Actually, you read my mind as I was about to suggest to close it (as you mentioned, this is fixed in v3). Thanks!

PR #503, opened 2020-08-04
Filter method with standard eloquent builder

marcoocram:
This PR enables the option to use the filters from a standard laravel query if you want to use the allowedfilters outside from the Spatie Query Builder.

It has no reason to use on this method the QueryBuilder instead of standard Builder (on filter interface uses the eloquent builder)


AlexVanderbist:
Thanks!

PR #505, opened 2020-08-05
Add elodo to front-end-implementation.md

MaximVanhove:
Adds elodo to the list of front-end implementations


AlexVanderbist:
Awesome, thanks!

PR #519, opened 2020-09-03
add ability to filter by nested relation scopes

ahmedsayedabdelsalam:
// users/1?filter[posts.older_than]=2020-09-03
in english: filter users who have posts older than exact time.

the older than scope is placed in the Post Model
I use the Query Builder on the User Model and I wanted to filter them with a scope placed in the related posts.

`AllowedFilter::scope('posts.older_than)`


ahmedsayedabdelsalam:
any plans adding this feature?

AlexVanderbist:
Neat, thanks for the PR!

PR #525, opened 2020-09-09
Add config param to allow disable InvalidFilterQuery exception

samuelhgf:
Hi folks!

This PR implements a configuration parameter that allows disable the InvalidFilterQuery exception.

I think that I haven't seen an error message when changing the filter parameter in any site until see in mine.

Any suggestions are welcome! Thanks!

@freekmurze you already accepted a PR from me, in the dropbox package. If you would like to see this one, will be very great!

### DIFF
--- a/src/Concerns/FiltersQuery.php
+++ b/src/Concerns/FiltersQuery.php
@@ -70,7 +70,7 @@ trait FiltersQuery
 
         $diff = $filterNames->diff($allowedFilterNames);
 
-        if ($diff->count()) {
+        if ($diff->count() && ! config('query_builder.disable_invalid_filter_query_exception')) {
             throw InvalidFilterQuery::filtersNotAllowed($diff, $allowedFilterNames);
         }
     }
--- a/config/query-builder.php
+++ b/config/query-builder.php
@@ -30,8 +30,8 @@ return [
     'count_suffix' => 'Count',
 
     /*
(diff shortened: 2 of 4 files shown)


aliene-ctrl:
Excellent!

src/Concerns/FiltersQuery.php
```diff
@@ -70,7 +70,7 @@ protected function ensureAllFiltersExist()
 
         $diff = $filterNames->diff($allowedFilterNames);
 
-        if ($diff->count()) {
+        if ($diff->count() && ! config('query_builder.disable_invalid_filter_query_exception')) {
```
dominikb, line 73: I would suggest splitting the check for the configuration flag into a separate if-statement with an early return at the beginning of this method.
This would make it easier to parse mentally and save some operations at runtime.

dominikb (changes requested):

AlexVanderbist:
Great PR, thanks! Ideally you'd capture the `InvalidQueryException` in your API's error handler and return a formatted error as JSON but I get where you're coming from. I'll add this info to the docs and give the whole section a page of its own. Thanks!

src/Concerns/FiltersQuery.php
AlexVanderbist: Good suggestion, I'll change it after merging :)

PR #532, opened 2020-09-24
Add LengthAwarePaginator to QueryBuilder

### LINKED ISSUE #531: whenAppended no longer working
I don't know if this is an QueryBuilder bug or Laravel has changed this call:

```
<?php

namespace App\Http\Resources;

use Illuminate\Http\Resources\Json\JsonResource;

class PostResource extends JsonResource
{
    /**
     * @return array
     */
    public function toArray($request)
    {
        return [
            'id' => $this->getRouteKey(),
            'thumbnail' => $this->whenAppended('thumbnail'),
        ];
    }
}
```

```
class IndexController extends Controller
{
    public function __invoke()
    {
        $models = QueryBuilder::for(Post::class)
            ->allowedAppends([
                'thumbnail',
            ])
            ->jsonPaginate();

        return PostResource::collection($models);
    }
}
```
The appending fields are just not included in the response.

francoism90:
When using `->paginate()` or `->jsonPaginate()`, it seems it doesn't inject the appends when needed.
This should fix #531.

This does work, but please let me know if the code can improved. :)

### DIFF
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -7,6 +7,7 @@ use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Http\Request;
+use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Traits\ForwardsCalls;
 use Spatie\QueryBuilder\Concerns\AddsFieldsToQuery;
@@ -111,6 +112,10 @@ class QueryBuilder implements ArrayAccess
 
         if ($result instanceof Collection) {
             $this->addAppendsToResults($result);
+        }
+
+        if ($result instanceof LengthAwarePaginator) {
+            $this->addAppendsToResults(collect($result->items()));
         }
 
         return $result;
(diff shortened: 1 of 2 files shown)


dominikb (commented):
Looks good to me. Does it make sense to add a test for this?

francoism90:
@dominikb Thanks for your review. :)

It would be great if this could be merged, this is the only showstopper for upgrading to L8.

francoism90:
@freekmurze Sorry, but will this ever be merged/fixed? It seems I'm not the only one having this issue and I'm only waiting for this fix or I need to downgrade the package to v2.

freekmurze:
@francoism90 please have a little patience. We'll handle this we have some time. There are a lot of other issues and client projects that require our attention as well.

@francoism90 can you add a test that is fixed by this PR. This will ensure that we don't break this behaviour again.

francoism90:
@freekmurze Sorry, I do understand client project have priority and I'm glad Spatie is maintaining so many open-source projects. :)

It doesn't seem to current `AppendTest` does tests only Model instances, no collections (this does work fine)/paginates. I'll try to rewrite the test, but it doesn't seem all that simple at first.

freekmurze:
Thanks for your work on this! 👍

francoism90:
@freekmurze Thanks for the merge! :+1:

PR #533, opened 2020-10-01
Update what defines an ignored filter value

### LINKED ISSUE #495: Zero ignored as a filter value
Hi,

First of all, thank you for this amazing package. It saved me a lot of time and it works really well. It has almost everything covered.

Still, there are some edge cases where it seems to fail.

In my app, I have a datatable of users and some filtering fields. One of them is a selector that lets you filter by "blocked status" to get only active, inactive(blocked) or both type of  accounts. The field in the DB is a Mysql Boolean.

The problem here is that when I send the value "0" it gets ignored and so the underlying query is not correctly formed.

I pinpointed where the problem is and I can fix it, but it would imply having to override the package through composer using PSR-4 tricks... an ugly solution.

The problem is in [FiltersPartial.php](https://github.com/spatie/laravel-query-builder/blob/master/src/Filters/FiltersPartial.php). There you are using array_filter with no parameters and so it assumes al false elements to be eliminated and that means 0 its considered false and eliminated. The solution would be to pass 'strlen' as a callback function as stated here [https://stackoverflow.com/a/27501297](https://stackoverflow.com/a/27501297).

This is not urgent as we can circumvent the problem using a custom filter or a scopeFilter, but I though I could share my experience and see if it can help others

Regards

dominikb:
Fixes #495 

As suggested the issue the addition of the `strlen` callback allows for falsy values (most often '0') to stay and only filter truly empty strings, like originally intended when this feature was implemented. (https://github.com/spatie/laravel-query-builder/commit/f7fd9a470890898f2814cf929a90e4770ea2d2c4)

### DIFF
--- a/src/Filters/FiltersPartial.php
+++ b/src/Filters/FiltersPartial.php
@@ -21,12 +21,12 @@ class FiltersPartial extends FiltersExact implements Filter
         $sql = "LOWER({$wrappedProperty}) LIKE ?";
 
         if (is_array($value)) {
-            if (count(array_filter($value)) === 0) {
+            if (count(array_filter($value, 'strlen')) === 0) {
                 return $query;
             }
 
             $query->where(function (Builder $query) use ($value, $sql) {
-                foreach (array_filter($value) as $partialValue) {
+                foreach (array_filter($value, 'strlen') as $partialValue) {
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

Issue #515, opened 2020-08-27
How can ignore the wrong filter prams ?


jeanali:
**New with this package**

I'm using custom query filter and have URL for filtering the data like this

http://example.test/posts?filter[name]=true

All things is good but once I try to change filter[name] to filter[anything] I got this error

 ` Requested filter(s) anything are not allowed. Allowed filter(s) `

How can check if it right do filter, failed show all posts like normal ?

aliene-ctrl:
This is actually a really annoying feature of an otherwise great package.

Is it just me or does [`ensureAllFiltersExist`](https://github.com/spatie/laravel-query-builder/blob/2258ce2d1accb5a286c9fef2bcf58398f1889dde/src/Concerns/FiltersQuery.php#L63) take it's concern a little too seriously? Is the exception really necessary? Is this method even necessary?

dominikb:
I think the exception does exactly what it is supposed to do: it let's you react to an invalid query by generating a meaningful error response. This is quite helpful if you do not control the consumer of your API.

@jeanali : what you were asking for could be solved like this

```php
try {
    $models = QueryBuilder::for(MyModel::class)->allowedFilters('name');
} catch (InvalidFilterQuery $exception) {
    // Maybe log this or extend the response to provide information about this fallback-behaviour to your API consumer
   $models = MyModel::all();
}
```

AlexVanderbist:
Hi! v3.3.0 adds a new config option to disable this exception. I've also opened a discussion for making the exception behave a bit better in a real world scenario here: #537 

Thanks for your concern and time!

PR #534, opened 2020-10-02
Remove untested and undocumented snake-casing of filter fields

### LINKED ISSUE #524: Bug on Select Field when field is write in UPPERCASE
when i try to get only PES_COD field, 

this error shows:
image

and if i get updated_at works fine..
this query works:

image

dominikb:
Fixes #524 

> > It seems that's a similar problem that occurs in [#515](https://github.com/spatie/laravel-query-builder/issues/515#issuecomment-689229711).
> 
> It's a really annoying feature. IMO `ensureAllFieldsExist` can be safely deleted.
> 
> To use this package I've had to wrap the `QueryBuilder` and override the `allowedFilters` method and comment out the line where it issues `$this->ensureAllFiltersExist();`

I think `ensureAllFieldsExist` is actually quite helpful and is not actually the problem for this issue.
The problem is the snake_casing of user-supplied parameters, as @marttosc already pointed out.

### DIFF
--- a/src/Concerns/AddsFieldsToQuery.php
+++ b/src/Concerns/AddsFieldsToQuery.php
@@ -73,8 +73,6 @@ trait AddsFieldsToQuery
             ->map(function ($fields, $model) {
                 $tableName = Str::snake(preg_replace('/-/', '_', $model));
 
-                $fields = array_map([Str::class, 'snake'], $fields);
-
                 return $this->prependFieldsWithTableName($fields, $tableName);
             })
             ->flatten()


marttosc:
In that case, I think these are two different issues.

1. The problem snake-casing problem really exists, as @dominikb also pointed out.
2. About the `ensureAllFiltersExists`, why not just ignore filters that aren't available?

dominikb:
> 2. About the `ensureAllFiltersExists`, why not just ignore filters that aren't available?

IMO it helps to provide a better user experience for consumers of your API. Silently ignoring errors can lead to unexpected behavior and frustrating, hard to debug errors. PR #525 provides a possible solution - let's continue the discussion there.

AlexVanderbist:
Thanks @dominikb, Another good one :) Definitely a breaking change tho so gonna tag this PR for the next major version.

Regarding `ensureAllFiltersExists`: catching the exception in the error handler and returning a proper JSON response is definitely the way to go (as mentioned in #525). Maybe making the exception `Responsable` is a good idea for the next version as well. Thanks all

PR #540, opened 2020-10-08
fix(filters): allow filters to be used without passing values

secmohammed:
removing this simple if statement turns the ability on to use scopes without having values, which can be  useful at certain scenarios, such as the following:
If I have a table that needs to filter on showing coupons that aren't expired yet. I don't need value as I have the current timestamp which I will be using. then I can pass at the query parameters the following 
```
host:port/endpoint?filter[without_expired_discounts]
```

### DIFF
--- a/src/AllowedFilter.php
+++ b/src/AllowedFilter.php
@@ -4,30 +4,41 @@ namespace Spatie\QueryBuilder;
 
 use Illuminate\Support\Collection;
 use Spatie\QueryBuilder\Filters\Filter;
-use Spatie\QueryBuilder\Filters\FiltersCallback;
 use Spatie\QueryBuilder\Filters\FiltersExact;
-use Spatie\QueryBuilder\Filters\FiltersPartial;
 use Spatie\QueryBuilder\Filters\FiltersScope;
+use Spatie\QueryBuilder\Filters\FiltersPartial;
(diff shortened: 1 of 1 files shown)


dominikb:
Hi @secmohammed,
thanks for contributing to this project.

I've checked out your pull request and ran the tests and there are a number of failing tests that would break current behavior. 
Applying filters without having submitted values would also render the `->ignore()` feature less useful.

I would suggest solving your problem with a custom or callback filter:
```php
// host:port/endpoint?filter[without_expired_discounts]=true

QueryBuilder::for(Model:class)
    ->allowedFilters(AllowedFilter::callback(
        'without_expired_discounts',
        function($query, bool $value, $property) {
            if (in_array($value, ['true', '1'])) {
                $query->withoutExpiredDiscounts();
            }
        }))
```

__As a tip for future pull requests:__
Try to limit proposed changes to one topic. You've included lots of restructuring in the file and new docblock comments, which in turn made the PR hard to review for the proposed change. (Allowing null values for filters.)

dominikb:
@secmohammed Could you edit your last response and add __how__ you worked around this issue? This could be helpful to others in the future. Thanks!

@AlexVanderbist This PR can be rejected and closed.

PR #542, opened 2020-10-11
Make nested scope compatible with older Laravel

hizzely:
Hi! this simple PR is a fix to:
```
BadMethodCallException: Method Illuminate\Support\Str::of does not exist.
```
Those who still use older version of Laravel (< 7) and upgraded the package from 3.2.4 to 3.3.0 might encounter this issue because the absent of Fluent String feature available starting from Laravel 7.


dominikb (approved):

freekmurze:
Thanks!

PR #545, opened 2020-10-14
implement default includes

CoolGoose:
This allows for default includes of relationships.

### DIFF
--- a/docs/features/including-relationships.md
+++ b/docs/features/including-relationships.md
@@ -29,6 +29,17 @@ $users = QueryBuilder::for(User::class)
 // $users will contain all users with their posts and permissions loaded
 ```
 
+You can load relatinships by default using:
+
+```php
+// GET /users?include=posts,permissions
+$users = QueryBuilder::for(User::class)
+    ->allowedIncludes([AllowedInclude::relationship('posts', null, true)])
+    ->get();
+
+// $users will contain all users with their posts and post counts
+```
+
 ## Disallowed includes
 
 When trying to include relationships that have not been allowed using `allowedIncludes()` an `InvalidIncludeQuery` exception will be thrown. Its exception message contains the allowed includes for reference.
--- a/src/AllowedInclude.php
+++ b/src/AllowedInclude.php
@@ -26,17 +26,17 @@ class AllowedInclude
         $this->internalName = $internalName ?? $this->name;
     }
 
-    public static function relationship(string $name, ?string $internalName = null): Collection
+    public static function relationship(string $name, ?string $internalName = null, bool $default = false): Collection
     {
         $internalName = Str::camel($internalName ?? $name);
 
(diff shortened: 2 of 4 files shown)


CoolGoose:
Definately not the prettiest thing, but it does the job, and i tried keeping as much backwords compatibility as possible with 'default' being the last parameter.

My use case is basically using it with laravel jetstream / inertiajs so not exactly a pure api, but I find it really usefull for default listings as well

dominikb:
Hi @CoolGoose,
thanks for contributing to this project.

Can you elaborate on your (and potentially other) use cases for this feature? As I understand this works a lot like [Eager Loading](https://laravel.com/docs/8.x/eloquent-relationships#eager-loading). If an `include` is defined as default it is always included in the response. Am I understanding this correctly?

Also, you might have noticed this project is very well tested and I think your PR has a higher chance of being merged if you add tests for it. If you need help with writing tests check out [existing test cases](https://github.com/spatie/laravel-query-builder/blob/master/tests/IncludeTest.phpl) to get you started.

CoolGoose:
Hi @dominikb 
Sorry for the late reply.

I'll add the tests for it, not sure why that slipped my mind 🤦‍♂️ 

Regarding the usage itself, I would say it's 'diverging' a bit from the normal API usage due to InertiaJS.
Case in point, usually on a let's say /posts endpoint from an UI perspective you would do something like /posts?include=comments as an AJAX call.

Since InertiaJS is sending the data as the props auto-magically, I am not doing a separate request by hand, so I am missing on some data that would be nice to have as default.

Again it's not a proper JSONAPI spec per se, but it helps in this case.

CoolGoose:
@dominikb I've added the test.

AlexVanderbist:
Hi @CoolGoose! Thanks for contributing. I think I'm still missing something: how are the following two examples different?

With a default relationship:
```php
    ->allowedIncludes([AllowedInclude::relationship('posts', null, true)])
    ->get();
// $users will contain all users and their posts
```

Or with an eager load:
```php
    ->with('posts')
    ->get();
// $users will contain all users and their posts
```
Can you elaborate?

CoolGoose:
@AlexVanderbist you will use the `posts_count` attribute.

AlexVanderbist:
Ah, I see! So what about the following?

```php
    ->with('posts')
    ->withCount('posts')
    ->get();
```

Trying to keep this package as simple as possible 😅

CoolGoose:
@AlexVanderbist that's fair. I thought it's more consistent to be able to set a default in both cases.
Can be closed if you think there's no added value for the project, thank you for the time reviewing this.

AlexVanderbist:
Hey, we've gone over this again and decided to add some documentation for these "default includes" using Laravel's `with` method. I'm really trying to limit the scope of this package to the absolute minimum to keep maintenance doable. Thank you for taking the time to PR this idea and I'm sorry it didn't get merged.

PR #547, opened 2020-10-16
Corrects the query-builder config name when calling it

samuelhgf:


samuelhgf:
@AlexVanderbist

src/Concerns/FiltersQuery.php
```diff
@@ -74,7 +74,7 @@ protected function ensureAllFiltersExist()
 
         $diff = $filterNames->diff($allowedFilterNames);
 
-        if ($diff->count()) {
+        if ($diff->count() && ! config('query_builder.disable_invalid_filter_query_exception')) {
```
AlexVanderbist, line 77: This isn't required anymore due to the boolean check above.

AlexVanderbist:
Right, thanks 😅

PR #550, opened 2020-10-19
Multidimensional values run scope multiple times

### LINKED ISSUE #530: Filtering ignored values if the supplied value is an array
Hello and thank you for plugin.

I found a problem when you using scope filter with multiple values.

Setup looks like:
```php
$query = QueryBuilder::for(Trade::class)
            ->allowedFilters([
                AllowedFilter::scope('durationBetween', 'durationBetween')->ignore(0, 'all'),
            ])
```

Scope:
```php
public function scopeDurationBetween(Builder $query, $from, $to): Builder
    {
        return $query->where('duration', '<=', (int)$from)
            ->where('duration', '>=', (int)$to);
    }
```

Query:
```php
Request URL: http://journal.test/api/trades?page=1&itemsPerPage=10&filter%5BdurationBetween%5D%5B0%5D=60000%2C1&filter%5BdurationBetween%5D%5B1%5D=300000%2C60000&filter%5BdurationBetween%5D%5B2%5D=900000%2C300000
```

Result
```json
{
    "message": "Array to string conversion",
    "exception": "ErrorException",
    "file": "/mnt/e/UbuntuServer/php/journal/vendor/spatie/laravel-query-builder/src/AllowedFilter.php",
    "line": 148,
    "trace": [
        {
            "function": "handleError",
            "class": "Illuminate\\Foundation\\Bootstrap\\HandleExceptions",
            "type": "->"
        },
        {
            "file": "/mnt/e/UbuntuServer/php/journal/vendor/spatie/laravel-query-builder/src/AllowedFilter.php",
            "line": 148,
            "function": "array_diff_assoc"
        },
        {
            "file": "/mnt/e/UbuntuServer/php/journal/vendor/spatie/laravel-query-builder/src/AllowedFilter.php",
            "line": 43,
            "function": "resolveValueForFiltering",
            "class": "Spatie\\QueryBuilder\\AllowedFilter",
            "type": "->"
        },
        {
            "file": "/mnt/e/UbuntuServer/php/journal/vendor/spatie/laravel-query-builder/src/Concerns/FiltersQuery.php",
            "line": 37,
            "function": "filter",
            "class": "Spatie\\QueryBuilder\\AllowedFilter",
            "type": "->"
        },
        {
            "file": "/mnt/e/UbuntuServer/php/journal/vendor/laravel/framework/src/Illuminate/Support/Traits/EnumeratesValues.php",
            "line": 202,
            "function": "Spatie\\QueryBuilder\\Concerns\\{closure}",
            "class": "Spatie\\QueryBuilder\\QueryBuilder",
            "type": "->"
        },
        {
            "file": "/mnt/e/UbuntuServer/php/journal/vendor/spatie/laravel-query-builder/src/Concerns/FiltersQuery.php",
            "line": 47,
            "function": "each",
            "class": "Illuminate\\Support\\Collection",
            "type": "->"
        },
        {
            "file": "/mnt/e/UbuntuServer/php/journal/vendor/spatie/laravel-query-builder/src/Concerns/FiltersQuery.php",
            "line": 27,
            "function": "addFiltersToQuery",
            "class": "Spatie\\QueryBuilder\\QueryBuilder",
            "type": "->"
        },
...
```

indykoning:
Fixes #530

Accept multidimensional values without crashing🔥 , use these values to run scope multiple times.
Before when a query such as `?filter[valueBetween][0]=2000,3000&filter[valueBetween][1]=3000,4000` was ran it would fail on "Array to string conversion" while checking for ignored values.

Now we jump deeper 🕳️  into the array to check if there are any ignored values.

If the values have been resolved we will run the scope however many times requested.

How the execution of the scope should be handled the second time should be defined in the scope.

### DIFF
--- a/src/AllowedFilter.php
+++ b/src/AllowedFilter.php
@@ -40,13 +40,19 @@ class AllowedFilter
 
     public function filter(QueryBuilder $query, $value)
     {
-        $valueToFilter = $this->resolveValueForFiltering($value);
+        $valuesToFilter = $this->resolveValueForFiltering($value);
 
-        if (is_null($valueToFilter)) {
+        if (is_null($valuesToFilter)) {
             return;
         }
 
-        ($this->filterClass)($query->getEloquentBuilder(), $valueToFilter, $this->internalName);
+        if (!isset($valuesToFilter[0]) || !is_array($valuesToFilter[0])) {
+            $valuesToFilter = [$valuesToFilter];
+        }
+
+        foreach ($valuesToFilter as $key => $valueToFilter) {
+            ($this->filterClass)($query->getEloquentBuilder(), $valueToFilter, $this->internalName);
+        }
     }
 
     public static function setFilterArrayValueDelimiter(string $delimiter = null): void
@@ -145,7 +151,19 @@ class AllowedFilter
     protected function resolveValueForFiltering($value)
     {
         if (is_array($value)) {
-            $remainingProperties = array_diff_assoc($value, $this->ignored->toArray());
+            $remainingProperties = [];
+            $multidimensional = false;
+            foreach ($value as $val) {
+                if (!is_array($val)) {
+                    continue;
+                }
+                $remainingProperties[] = $this->resolveValueForFiltering($val);
+                $multidimensional = true;
+            }
+
+            if (!$multidimensional) {
+                $remainingProperties = array_diff_assoc($value, $this->ignored->toArray());
+            }
 
             return ! empty($remainingProperties) ? $remainingProperties : null;
         }
--- a/tests/FilterTest.php
+++ b/tests/FilterTest.php
@@ -361,6 +361,24 @@ class FilterTest extends TestCase
     }
 
     /** @test */
+    public function it_can_filter_by_scope_multiple_times()
+    {
+        Carbon::setTestNow(Carbon::parse('2016-05-05'));
+
+        $testModel = TestModel::create(['name' => 'John Testing Doe']);
+
+        $modelsResult = $this
+            ->createQueryFromFilterRequest(['created_between' => [
+                ['2016-01-01', '2017-01-01'],
+                ['2015-01-01', '2018-01-01'],
+            ]])
+            ->allowedFilters(AllowedFilter::scope('created_between'))
+            ->get();
+
+        $this->assertCount(1, $modelsResult);
+    }
+
+    /** @test */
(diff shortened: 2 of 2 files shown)


AlexVanderbist:
Hey, I've been thinking about this for a while and I think it comes down the whether we want to support multiple of the same filter in one request or not. 

For example: `?filter[valueBetween][0]=2000,3000&filter[valueBetween][1]=3000,4000` could mean "use the `valueBetween` filter twice", which makes sense for adding a scope twice. However, for a custom filter it might make more sense to pass the entire request array to the custom `valueBetween` filter once. Like this: `[[2000, 3000], [3000, 4000]]`.

As I said, in the context of scopes, doing it like your example makes a lot of sense. However, I think I'd rather not add too much magic or config options to the filters and keep the "only one filter of every type" rule. What do you think?

AlexVanderbist:
Tagging this PR as "next major" as it's definitely gonna be a breaking change

freestyledork:
I vote for support of this feature :) 👍

indykoning:
Passing the whole request array would probably be the solution where the programmer gets most control out of the request data.
However this would be the biggest breaking change, since every scope and filter would have to be changed.
Also scopes used outside of this module may have to be built to support an array as first argument or regular values so they can still be used as normal.
```
public function valueBetween($value1, $value2 = null) {
    if (!is_array($value1)) {
        $value1 = [$value1, $value2]
    }

    foreach($value1 as $data) {
...
```

I know we don't want to add too many configuration options, however in this case i do think it would be best to add a switch in the `AllowedFilter` class such as `passFullRequest(true|false)`, default functionality if not called would be to call the function multiple times, in order to break code as little as possible. calling `passFullRequest()` will cause it to pass the full request data for that filter unless it has false passed as argument.

Example usages would be
```
// Not really needed to add the passFullRequest
->allowedFilters([
    AllowedFilter::scope('durationBetween', 'durationBetween')->ignore(0, 'all')->passFullRequest(false),
])
```
And
```
->allowedFilters(AllowedFilter::callback('durationBetween', function (Builder $query, $value, string $_property) {
        foreach($value as $range) {
            $query->orWhere(function($query) use ($range) {
                [$min, $max] = $range;

                $query->where('min', '>=', $min);
                $query->where('max', '<=', $max);
            });
        }
    }))->passFullRequest()
```

This way someone will be able to pick their preference for its functionality.

PR #567, opened 2020-11-23
Fix to prevent ambiguity sql errors 

### LINKED ISSUE #566: Ambiguous filters partial
I noticed that a PR was merged to fix the issue that you can get an ambiguity sql error when filter and sorting at the same time when the table sorted on is joined and has a overlapping field name.

For consistency sake I will make a similar PR to fix the same problem for Partial Filters.

jonneroelofs:
fixes #566


AlexVanderbist:
Perfect, thanks!

PR #585, opened 2021-01-04
Fix the defaultDirection example

waltoncon:
In the third custom sorts example, `defaultDirection` is shown as chaining from `QueryBuilder`/`SortsQuery` but is actually a method on `AllowedSort`.


GersonM17 (approved):

freekmurze:
Thanks!

Issue #345, opened 2019-09-26
QueryBuilderRequest only using $this->query(), maybe change to $this->input()?


Haskiez:
Maybe this is a bit narrow in focus but, it would be nice to use the QueryBuilder for querying your data but also for updating it. This package is great if you are putting everything in the query string, but what if you want to use this QueryBuilder to handle scoping for something like updating via a PUT request?

Let's say we want to update specific records that match the filters being passed. Since the package only uses $this->query() it won't pick up anything in the body. It would be nice if it used $this->input() instead so it could find the filters, includes, and fields in both the body and in the query string.

Essentially the only change would be updating it to use the right ->input() instead of ->query() in the QueryBuilderRequest file.

AlexVanderbist:
How exactly would you be using it with a PUT request? I think it's actually more appropriate to keep the filters and data separated but maybe I'm missing something?

Tiny untested example:
```php
QueryBuilder::for(User::class)
->allowedFilters(['email'])
->get()
->update($request->input());
```
Would work for the following request: `PUT /users?filter[email]=gmail` and regular update data as the request body.

Haskiez:
@AlexVanderbist maybe it does. Maybe its more personal preference that when making PUT or POST requests, it feels funky adding things to the url rather than putting it in the body. But I see your point and that also does make sense.

AlexVanderbist:
I suppose if you really wanted something like this, you can extend the `QueryBuilderRequest` class, make the necessary changes and bind it in the service container... Gonna close this for now, feel free to open a new issue if you're still experiencing any issues

PR #589, opened 2021-01-20
Allow to get params from $request->input

arrilot:
Our Api Design Guide forces us to follow the following convention for search requests:
```
POST /api/v1/customers:search
{
   "include": ["addresses"],
   "sort": ["-created_at,"id"]
   ....
}
```
It makes some sense because we can avoid some limitations of get requests this way like "request string is too long"

`laravel-query-builder` can help us a lot, but there is a problem - it always gets datafrom $request->query() no matter of request method and we can't move those fields to query string due to Design Guide.

The current solutions that I'm aware of are not great.
1. Pass `new Request()` to *every* query builder call like that $query = QueryBuilder::for(Customer::class, new Request($request->all())) 
This seems hacky and not very convenient, especially if you call query builder in a scope without `$request` available

2. 
   - inherit `QueryBuilderRequest` (you actually need to copy paste the majority of it to the child class btw), 
   - replace `QueryBuilderRequest` binding in the service container with the new one.
   - inherit `QueryBuilder` itself because it still creates old `QueryBuilderRequest` due [this ](https://github.com/spatie/laravel-query-builder/blob/master/src/QueryBuilder.php#L45) and [that](https://github.com/spatie/laravel-query-builder/blob/master/src/QueryBuilder.php#L68)
  - start using the inherited QueryBuilder everywhere.
  Frankly, this looks extremely fragile. I'm not even sure it 100% works now, and I'm even less sure it won't break in the future package releases.

So, the idea is to add this kind of stuff to the package itself, especially since the issue was already raised before by some ppl, e.g #345

It does not seem to be a good idea to blindly change everything from $request->query() to $request->input() or smth else because it will definitely cause BC breaks.

That's why this PR introduces new `inspect_input_instead_of_query => false|true` configuration option

If the developer changes it to `true` the packages starts to seek `include`, `sort` e.t.c parameters in `$request->input()` instead of `$request->query()`

### DIFF
--- a/config/query-builder.php
+++ b/config/query-builder.php
@@ -35,4 +35,9 @@ return [
      */
     'disable_invalid_filter_query_exception' => false,
 
+    /*
+     * By default the package inspects query string of request - $request->query().
+     * You can change this behavior to $request->input() by setting this value to `true`.
+     */
+    'inspect_input_instead_of_query' => false,
 ];
--- a/src/QueryBuilderRequest.php
+++ b/src/QueryBuilderRequest.php
@@ -36,10 +36,10 @@ class QueryBuilderRequest extends Request
     {
         $includeParameterName = config('query-builder.parameters.include');
 
-        $includeParts = $this->query($includeParameterName);
+        $includeParts = $this->getInput($includeParameterName);
 
         if (! is_array($includeParts)) {
-            $includeParts = explode(static::getIncludesArrayValueDelimiter(), $this->query($includeParameterName));
+            $includeParts = explode(static::getIncludesArrayValueDelimiter(), $this->getInput($includeParameterName));
         }
 
         return collect($includeParts)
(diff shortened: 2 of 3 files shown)


AlexVanderbist:
Neat! Thanks for the tested PR and sorry for the delay. I'll merge this but change the config key to `request_data_source` with value `body` or `query_string` or something. Thanks!

stevebauman:
Thanks for the recent merges and release @AlexVanderbist!! Much appreciated!

arrilot:
@AlexVanderbist I think `body` is a little missleading because for GET requests `$request->input()` operates with query string instead of request body 

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Http/Concerns/InteractsWithInput.php#L265
https://github.com/laravel/framework/blob/8.x/src/Illuminate/Http/Concerns/InteractsWithInput.php#L265

Issue #632, opened 2021-05-24
Using 'use_simple_pagination' breaks appends


francoism90:
I was wondering why my appends wouldn't work, turns out using `'use_simple_pagination' => true,` causes this to happen. Is this is a know limitation? Can it be fixed? :)

Thanks.

francoism90:
Wrong repo.

freekmurze:
Thanks for the PR

PR #633, opened 2021-05-24
Fix simple paginator append not working

francoism90:
Laravel offers more pagination like the newly introduced cursor. I didn't test that one, but may be interested to add as well. :)

For now this should fix when someone uses a simple paginate which is not length aware #632 .


freekmurze:
Thanks!

PR #637, opened 2021-06-08
fix unexpected lowecasing of appends

demoskp:
Appends from the request get lowercased which results in non-matching request appends and allowedAppends.

i.e. `allowedAppends('someComelCaseAppend')`
`/api-url/endpoint?append=someComelCaseAppend`

This causes the request parameter append with value `someComelCaseAppend` to become `somecamelcaseappend` which is not the expected behaviour.


AlexVanderbist:
Perfect, thanks!

Issue #640, opened 2021-06-22
Filter not recognizing relationship


mitchierichie:
https://spatie.be/docs/laravel-query-builder/v3/features/filtering#exact-or-partial-filters-for-related-properties

Following the instructions in the link above, I tried to create an AllowedFilter for one of my relationships. My query fails because instead of filtering by my related model, it tries to find a column with that name on the main model's table.

```log
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'product_model.slug' in 'where clause' 
```

 I did some digging - seems Spatie\QueryBuilder\Filters\FiltersExact, L51 is responsible. I've included the source for that method here:

```php
    protected function isRelationProperty(Builder $query, string $property): bool
    {
        if (!Str::contains($property, '.')) {
            return false;
        }

        if (in_array($property, $this->relationConstraints)) {
            return false;
        }

// line 51
        $firstRelationship = Str::camel(explode('.', $property)[0]);
        
        if (!method_exists($query->getModel(), $firstRelationship)) {
            return false;
        }

        return is_a($query->getModel()->{$firstRelationship}(), Relation::class);
    }

```

Why would it change the requested relationship name to camel case? Wouldn't you just use the exact relationship name provided by the developer, not modify it and then fail to find relationships?

AlexVanderbist:
>Why would it change the requested relationship name to camel case? Wouldn't you just use the exact relationship name provided by the developer, not modify it and then fail to find relationships?

This tried to follow 2 best practices: case insensitive urls & camelcase for Laravel relationships. Ultimately not the best choice for this package. Let's continue this issue in the #641

mitchierichie:
@AlexVanderbist I see, well I appreciate the quick turnaround

PR #641, opened 2021-06-23
Use exact relationship provided by developer

mitchierichie:
Changing the relationship method name provided by the developer to camel case is bad practice. This is also undocumented and unexpected behavior and quite frustrating as a developer using this package. The name of the filter is exact - it should be _exact_.

### DIFF
--- a/src/Filters/FiltersExact.php
+++ b/src/Filters/FiltersExact.php
@@ -48,7 +48,7 @@ class FiltersExact implements Filter
             return false;
         }
 
-        $firstRelationship = Str::camel(explode('.', $property)[0]);
+        $firstRelationship = explode('.', $property)[0];
 
         if (! method_exists($query->getModel(), $firstRelationship)) {
             return false;


mitchierichie:
Fixes #640

mitchierichie:
@AlexVanderbist I think I missed a few places where the camel case conversion comes in but hopefully my idea is clear. I will try to update this PR soon with a full fix for my mentioned issue #640 but I don't know this codebase in and out...

should I close this PR until I can add more commits?

AlexVanderbist:
Thanks, we've been pushing back this change for a while because it's a breaking change. I'll merge this into master, go over the rest of the package and tag v4 of the package later today :)

AlexVanderbist:
I've prepared the rest of the package to no longer camel-case any values for filters, includes and fields. It's available on the master branch and using `composer require spatie/laravel-query-builder@dev-master`. I'll update the docs later before tagging v4

mitchierichie:
@AlexVanderbist thank you!

PR #653, opened 2021-08-20
Add pagination methods with appended query to next / previous URLs

d8vjork:
So when we use cursor pagination on our Laravel projects they doesn't maintain the filters, sorts, appends, etc.

This will return previous/next URLs on the cursorPagination with all of them appended so next URLs still returning relevant results.


d8vjork:
Hope this makes sense for you guys, I can also provide some extra tests for make sure that this isn't broken in the future

AlexVanderbist:
Hi, thanks for the PR! An extra test would be helpful to make sure the changes don't break anything unexpected and everything keeps working in the future. Do you think we need something similar for the regulation pagination too?

d8vjork:
@AlexVanderbist most probably, will create tests & the rest of them asap

d8vjork:
All right, although there are other tests failing (non related to this I think)he implementation is a little bit crappy

d8vjork:
Well don't know how to test on PestPHP as I've never used it

I'll try to update this with v5's as soon as I've free time

freekmurze:
@d8vjork https://testing-laravel.com for learning how to use Pest 🙂

d8vjork:
@freekmurze So there's a way to progressively migrate tests from PHPUnit as Pest, I'll try with some of my projects. Thanks

PR #678, opened 2021-10-24
Extend FiltersPartial to support AND (+) and NOT (-)

patrickrobrecht:
Currently the `FiltersPartial` supports by filtering by multiple `field=a,b`. Those leads to a query `field LIKE "%a%" OR field LIKE "%b%"`. This PR extends the filter to support logical AND and NOT as well (and provides tests for the extended functionality via the new `FiltersComplexPartialTest` class). 

By default, the new feature is not enabled for `AllowedFilter::partial` (although setting the last parameters to something else than null would enable them). The new `AllowedFilter::complexPartial` enables the new filtering options with `+` for AND and `-` for NOT (as known from the input field of search engines). The + / - can be replaced with other strings by changing the default values when calling `AllowedFilter::complexPartial`.

Examples:
`field=a+b`
=> `field LIKE "%a%" AND field LIKE "%b%"`

`field=-a`
=> `field NOT LIKE "%a%"`

`field=a + -b`
=> `field LIKE "%a%" AND field NOT LIKE "%b%"`

`field=a+b, c`
=> `(field LIKE "%a%" AND field LIKE "%b%") OR field LIKE "%c%"`

Both AND separator and the prefix for NOT can be configured (they default to `+` and `-` if the filter is created via `AllowedFilter::complexPartial`).

`AllowedFilter::partial` results in the same query results as before. There is just one small difference regarding the generated SQL: In the case of an array input with just one element (as in the `falsy_values_are_not_ignored_when_applying_a_partial_filter` test) the new implementation avoids the unnecessary nesting in the query. (I discovered this because this test failed in the first run after my changes due to an additional nesting level leading to `select * from test_models where ((LOWER(test_models.id) LIKE ?))` - when fixing this I removed both unnecessary levels such that the `()` are removed.)

### DIFF
--- a/.github/workflows/update-changelog.yml
+++ b/.github/workflows/update-changelog.yml
@@ -1,0 +1,28 @@
+name: "Update Changelog"
+
+on:
+  release:
+    types: [released]
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: main
+
+      - name: Update Changelog
+        uses: stefanzweifel/changelog-updater-action@v1
+        with:
+          latest-version: ${{ github.event.release.name }}
+          release-notes: ${{ github.event.release.body }}
+
+      - name: Commit updated CHANGELOG
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          branch: main
+          commit_message: Update CHANGELOG
+          file_pattern: CHANGELOG.md
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 All notable changes to `laravel-query-builder` will be documented in this file
 
-## 4.0.0 - unreleased
+# 4.0.1 - 2021-10-27
+
+- revert deferred service provider (#677)
+
+## 4.0.0 - 2021-10-20
 
 - nested filters will no longer be automatically camel-cased to match a relationship name
 - includes will no longer be automatically camel-cased to match a relationship name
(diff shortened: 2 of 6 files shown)


patrickrobrecht:
Hi @AlexVanderbist,

would this feature be interesting for the next laravel-query-builder release? In my opinion, implementation and tests are complete, just the updates of the documentation is missing.

jpgilchrist:
@patrickrobrecht if not part of the core package, perhaps you could release a package with this custom filter type. We could then include it in our AllowedFilters array.

patrickrobrecht:
@jpgilchrist Of course releasing an own package would be possible. However I would prefer an integration into the core package to avoid the overhead of maintaining a separate package with just a few lines.

jpgilchrist:
I totally understand that desire. However, when incorporating into the core package it has to be how everyone would want it, which makes things more complicated.

I would imagine this doesn't quite align with the goals. 

For instance almost everyone needs "search" but they don't have a built in search and instead they have custom filter or callback. 

Almost everyone needs a less than or greater than, but that's also not built in. 

One thing we did on a project was to extend query builder to make functions like 

```
CustomQueryBuilder(Review::class)
  ->withLessThan('rating')
```

That way it was always available. I'm not trying to dismiss the idea. I think it's a good one, just not sure it's a perfect fit for the core. That being said, if they implemented it, I'd probably use it.

jolora:
This looks brilliant and is exactly the feature I was just looking for. Did you make a decision on whether you would release your work as a package @patrickrobrecht ? It would be greatly appreciated by me at least 😊

For the record, I would also prefer to see this in core.

patrickrobrecht:
@jolora Thanks for the feedback. I currently have no plans for releasing a package, I may reconsider the release of an extension package if I have more than just this extended filter to provide in the extension.

In case of positive feedback by package maintainers @AlexVanderbist and @freekmurze, I would gladly offer to adapt the tests.

PR #682, opened 2021-11-02
Allows for selecting fields of related models with custom names

robertmarney:
This allows for a relationship to be defined which varies from the underlying table and does not require the user to know the underlying table name.

For example relation of user() but uses a table of private_users would fail in the old paradigm.


freekmurze:
Could you add a test for this?

robertmarney:
@freekmurze  - Yes I could, It might be a few days before I find some time.

PR #684, opened 2021-11-09
Use static instead of self

geidelguerra:
This will improve auto completion in vscode.


madrussa:
This would break PHP Compatibility with PHP 7.1 as listed on the requirements. Static return types were introduced in PHP 8 https://wiki.php.net/rfc/static_return_type. This would be a breaking change for anyone still on 7.1 and would usually require a major version increment. 
https://spatie.be/docs/laravel-query-builder/v3/requirements

freekmurze:
Thanks! Going to leave things as they are for now.

PR #702, opened 2021-12-24
PHP 8.1 Support

Medalink:
Trying out this package on PHP 8.1 fails with `During inheritance of ArrayAccess: Uncaught ErrorException: Return type of Spatie\QueryBuilder\QueryBuilder::offsetSet($offset, $value) should either be compatible with ArrayAccess::offsetSet(mixed $offset, mixed $value): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /app/vendor/spatie/laravel-query-builder/src/QueryBuilder.php:168`

This PR addresses the issue by adding `#[\ReturnTypeWillChange]` to the offset methods.

I also ran into an issue with null being passed to explode: `explode(): Passing null to parameter #2 ($string) of type string is deprecated {"exception":"[object] (ErrorException(code: 0): explode(): Passing null to parameter #2 ($string) of type string is deprecated at /app/vendor/spatie/laravel-query-builder/src/QueryBuilderRequest.php:55)`

I fixed it by adding an extra null check:

https://github.com/Medalink/laravel-query-builder/blob/67ad6d786f34717a14b5d5fcc77735c86a18bfa9/src/QueryBuilderRequest.php#L54-L56

### DIFF
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -153,21 +153,25 @@ class QueryBuilder implements ArrayAccess
         $this->subject->{$name} = $value;
     }
 
+    #[\ReturnTypeWillChange]
     public function offsetExists($offset)
     {
         return isset($this->subject[$offset]);
     }
 
+    #[\ReturnTypeWillChange]
     public function offsetGet($offset)
     {
         return $this->subject[$offset];
     }
 
+    #[\ReturnTypeWillChange]
     public function offsetSet($offset, $value)
     {
         $this->subject[$offset] = $value;
     }
 
+    #[\ReturnTypeWillChange]
     public function offsetUnset($offset)
     {
         unset($this->subject[$offset]);
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -153,26 +153,22 @@ class QueryBuilder implements ArrayAccess
         $this->subject->{$name} = $value;
     }
 
-    #[\ReturnTypeWillChange]
-    public function offsetExists($offset)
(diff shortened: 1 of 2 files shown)


freekmurze:
> This PR addresses the issue by adding #[\ReturnTypeWillChange] to the offset methods.

Thanks! Could you fix this by adding `: void` instead of that attribute?

Medalink:
@freekmurze using bool for the offsets and void for the other two work just as well. I was not sure how that would work with backwards compatibility but I updated the PR with return types that fix this without the need for `#[\ReturnTypeWillChange]`

freekmurze:
Thanks!

ankurk91:
@freekmurze 

mailcoach is using v3 of this package and we are getting same error while running on 8.1 :(

can you guys patch v3.0 as well?

freekmurze:
We're not actively working on 3.0 anymore. If you need a fix for this, feel free to PR the desired changes and we'll consider merging it in.

src/QueryBuilder.php
```diff
@@ -153,22 +153,22 @@ public function __set($name, $value)
...
     }
 
-    public function offsetExists($offset)
+    public function offsetExists($offset): bool
     {
         return isset($this->subject[$offset]);
     }
 
-    public function offsetGet($offset)
+    public function offsetGet($offset): bool
```
ankurk91, line 161: offsetGet should return `mixed` (Can return all value types) type

https://www.php.net/manual/en/arrayaccess.offsetget.php

PR #709, opened 2022-01-09
[3.0] fix: php 8.1 support

ankurk91:
backport https://github.com/spatie/laravel-query-builder/pull/702 
to v3.x


freekmurze:
Please fix this by implementing the typehints instead of the attributes.

ankurk91:
Version 3 need to support php v7.3 as well, the typehints would work on php 7.3?

ankurk91:
type-hint worked on php 7.3.

I have replaced the attributes with types, please check.

freekmurze:
Thanks!

PR #713, opened 2022-01-13
Update README.md

wayz9:
Replacing v3 to v5 in the urls so they can work and point to latest version of docs. 😎


freekmurze:
Thanks!

PR #724, opened 2022-02-08
V4 - Dynamic parameter names

ManukMinasyan:
Hi guys this is my first contribution to Spatie.
I will be glad to hear your opinions.
Thank you.

Usage example:

```php
QueryBuilder::for($feedback->comments(), null, [
    'sort' => 'discussionSort'
])
```


freekmurze:
Could you also update the readme and add tests. Maybe also provide a little bit of context why you would need this feature.

saibotk:
This is super useful for pages with multiple tables, we also have many use cases and would love to see this feature.

Issue #448, opened 2020-03-14
add custom query for include


MoeCasts:
For example, if just wanna take 10 comments of posts and order by created_at desc, can do like this.

```php
$posts = QueryBuilder::for(App\Models\Post::class)
            ->allowedIncludes([
                'comments' => function ($query) {
                    $query->orderBy('created_at', 'desc')->take(10);
                }
            ]);
```

If we can do like this, it would be better.

AlexVanderbist:
The easy solution for that would be to  create new relationship on the `Post` model that has the `orderBy` and `take` in there:

```php
public function highlightedComments() {
    return $this->comments()->orderBy('created_at', 'desc')->take(10);
}
```
And use it: `->allowedIncludes(['highlightedComments'])`.

Alternatively I'd accept a PR that adds support for custom include classes.

Good luck!

Yeenot:
@AlexVanderbist but what about if I want to sort the result of the `->allowedIncludes`?

ajcastro:
You can use 
```php
AllowedInclude::custom('comments', new IncludeWithQuery(function ($query) { $query->where(...); }))
```
```php
<?php

namespace App\Support\QueryBuilder;

use Illuminate\Database\Eloquent\Builder;
use Spatie\QueryBuilder\Includes\IncludeInterface;

class IncludeWithQuery implements IncludeInterface
{
    protected $callback;

    public function __construct(callable $callback)
    {
        $this->callback = $callback;
    }

    public function __invoke(Builder $query, string $include)
    {
        $callback = $this->callback;

        return $query->with([
            $include => $callback
        ]);
    }
}
```

PR #726, opened 2022-02-11
add include with callback feature

hsmfawaz:
**Greetings, Spatie team**

This PR intended to enhance includes feature with a callback.

More info #448


yusufonur:
Very good. 

Whenever we had to load a relation, we had to create a separate relation within the model or deviate by other side paths. 

When the whole relationship is loaded, situations may occur trying to get millions of records and uses such as limit or pagination may be required.

PR #742, opened 2022-03-17
Fix warning from passing null to explode for includeParts

### LINKED ISSUE #741: PHP8.1 passing null to explode() is deprecated warning for $includeParts
Identical to PR #702 ([commit](https://github.com/spatie/laravel-query-builder/pull/702/commits/67ad6d786f34717a14b5d5fcc77735c86a18bfa9)), the `include` parameter on a QueryBuilderRequest can be `null` and should be checked before passing to the `explode()` function.

https://github.com/spatie/laravel-query-builder/blob/e7abb43f1ee94753a2be3b4dc6e79eaf03d4a9d5/src/QueryBuilderRequest.php#L41

Update to `if (! is_array($includeParts) && ! is_null($includeParts)) {`

Currently bogged with other work but can try to make a PR soon if someone doesn't beat me to it.

steven-fox:
Closes #741

### DIFF
--- a/src/QueryBuilderRequest.php
+++ b/src/QueryBuilderRequest.php
@@ -38,8 +38,8 @@ class QueryBuilderRequest extends Request
 
         $includeParts = $this->getRequestData($includeParameterName);
 
-        if (! is_array($includeParts)) {
-            $includeParts = explode(static::getIncludesArrayValueDelimiter(), $this->getRequestData($includeParameterName));
+        if (! is_array($includeParts) && ! is_null($includeParts)) {
+            $includeParts = explode(static::getIncludesArrayValueDelimiter(), $includeParts);
         }
 
         return collect($includeParts)->filter();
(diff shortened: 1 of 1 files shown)


src/QueryBuilderRequest.php
```diff
@@ -38,8 +38,8 @@ public function includes(): Collection
 
         $includeParts = $this->getRequestData($includeParameterName);
 
-        if (! is_array($includeParts)) {
-            $includeParts = explode(static::getIncludesArrayValueDelimiter(), $this->getRequestData($includeParameterName));
+        if (! is_array($includeParts) && ! is_null($includeParts)) {
```
dominikb, line 41: Maybe you could change this to `is_string($includeParts)` and eliminate the other checks. I think this would avoid the warning as well as improve readability.

steven-fox: Agreed. Just saw this is the conditional on line 82. 👍

dominikb (approved):

PR #749, opened 2022-04-24
Fix Laravel 9 PHPStan generic check 

kayw-geek:
PHPStan error : Method App\Http\Controllers\Resource\DemoController::withQuery() has parameter $query with generic class Illuminate\Database\Eloquent\Builder but does not specify its types: TModelClass


freekmurze:
Thanks!

PR #775, opened 2022-06-06
static return type when returning $this

lorenzolosa:
I've changed the return types of some chainable methods in QueryBuilder from `self` to `static`, as it's more precise.

I've noticed this as I'm extending the QueryBuilder class with an additional method.


freekmurze:
Thanks!

PR #781, opened 2022-06-27
Fix Laravel 9 PHPStan generic check for  `__invoke()` method of Filter

kayw-geek:
Fix `PHPStan` error of generic check  when `FiltersXXXClass` used for extends.


freekmurze:
Thanks!

Issue #792, opened 2022-07-23
Include Relation is Null


lrfahmi:
I am confused by the results that do not match what is documented. I'am using the package version 5.0.2.
```
Package Version
{
    "name": "spatie/laravel-query-builder",
    "version": "5.0.2",
    ...
}
```

```
App\Models\Post.php
...

protected $fillable = [
	'user_id', 'title', 'body',
];

public function author()
{
	return $this->belongsTo('App\Models\User', 'user_id');
}
```

```
App\Http\Resources\PostResource.php
...

class PostResource extends JsonResource
{
    public function toArray($request)
    {
        return parent::toArray($request);
    }
}
```

```
App\Http\Controllers\Api\PostController.php
...

class PostController extends Controller
{
    public function index()
    {
        $posts = QueryBuilder::for(Post::class)
            ->allowedFields([
                'id', 'title', 'body',
                'author.id', 'author.name', 'author.email',
            ])
            ->allowedIncludes(['author'])
            ->jsonPaginate();

        return PostResource::collection($posts);
    }
}

```

Based on the information above, the response from the API is:

```
http://laravel.local:8083/api/posts?include=author&fields[posts]=id,body&fields[author]=id,name,email

{
    "data": [
        {
            "id": 1,
            "body": "Lorem ipsum dolor sit ame.",
            "author": null
        }
    ],

    ...
```

Please help, what's wrong? I redo the documentation but the result is the same. Thank you.

lrfahmi:
Some time later I tried with a different URL, entering the **user_id** to **fields[posts]**.
```
http://laravel.local:8083/api/posts?fields[posts]=id,user_id,body&fields[author]=id,name&include=author
...

Response: {
    "data": [
        {
            "id": 1,
            "user_id": 1,
            "body": "Lorem ipsum dolor sit ame.",
            "author": {
                "id": 1,
                "name": "Fahmi Ramdhani",
                "email": "fahmi@laravel.local",
                "email_verified_at": null,
                "created_at": "2022-07-23T02:41:07.000000Z",
                "updated_at": "2022-07-23T02:41:07.000000Z"
            }
        }
    ],
```

The select feature does not match the requested parameters, the author field is displayed all.

lrfahmi:
Apparently, the problem is from the fields[relation?] parameter. Where the query builder package or laravel specify the name / alias plural. So there is a difference in the Controller.

```
App\Models\Post.php
...

public function author()
{
[...]

Issue #681, opened 2021-11-01
AllowedFields are not taking into consideration the relationship fields


PharaohAsh:
**Using V4.0.1**

When we do the following:
```
QueryBuilder::for(Post::class)
    ->allowedFields('author.id', 'author.name')
    ->allowedIncludes('author');
```

And query the endpoint like so:
```
GET /posts?include=author&fields[author]=id,name
```

It appears to list _all_ the fields of the `author` and not **just** `id` and `name`.

I think the issue lies [here `->get($table)`](https://github.com/spatie/laravel-query-builder/blob/main/src/Concerns/AddsFieldsToQuery.php#L57) where it's grabbing the `$table` instead of the `$relation` from the collection of fields.

wattsie:
I too have just ran into the similar issue (v4.0.1 as well)

I tried applying a resource to each of the keys within the main resource. ie.
```
    public function toArray($request)
    {
        // return parent::toArray($request);
        return [
            'id' => $this->id,
            'name' => $this->name,
            'first_include' => (new FirstResource($this->first)),
            'second_include' => (new SecondCollection($this->second)),
        ];
    }
```
But it appears, when I call the `$this->first` it pulls in the entire record, thus does not obey the include'd requirement within the URL.

It would be awesome if any included/related models apply their resource or resource collections.

wattsie:
FYI: I have found a work around.

By checking the `relationsToArray` I have found which additional includes have been defined.

Then I just check and convert into either Resource or ResourceCollection as required.

```
class PrimaryResource extends JsonResource
{
    public function toArray($request)
    {
        $ret_val = [
            'id' => $this->id,
            'name' => $this->name,
        ];
        // Include resource
        if (array_key_exists('first_include', $this->relationsToArray())) {
            $ret_val['first_include'] = (new FirstIncludeResource($this->first_include));
        }
        // Include a resource collection
        if (array_key_exists('second_include', $this->relationsToArray())) {
            $ret_val['second_include'] = (new SecondIncludeCollection($this->second_include));
        }
        return $ret_val;
    }
}
```

jpgilchrist:
I just ran into this issue as well today. Same thing specifying fields on a relationship is not added to the select statement. If the field is for the root object, then it does work. 

?include=author&fields[post]=id&fields[author]=id,name

outputs 

`select post.id from post ... `

omitting the ?field[post]=id just results in a `select * ...`

claudiofabiao:
[...]

Issue #440, opened 2020-03-03
Remove CamelCasing relationship names


dominikb:
This is a follow-up ticket for issue #393. (and #366, #332, #307)

Although I personally use camelCased relationship names, I don't think this is a behavior we want to enforce. Especially because there have been multiple tickets regarding this feature.

The conversion happens in two places. Namely in [AddsFieldsToQuery](https://github.com/spatie/laravel-query-builder/blob/95cffd71f64e90ff9102481fdfb12f243d5425db/src/Concerns/AddsFieldsToQuery.php#L54) and in [AllowedInclude](https://github.com/spatie/laravel-query-builder/blob/44aee49b9edf03587bb59d0ee8041eae6e5d2304/src/AllowedInclude.php#L31).

I see multiple ways to handle removing the camel casing:
1. **Removing** the feature with a notice in the changelog.
2. **Remove** the feature, detect if a cast would be necessary and throw a **meaningful exception**.
3. Keep the camel casing and **dynamically** detect if a **conversion** is required.

What are your thoughts?

CC: @suth

suth:
Although it's documented the behavior surprised me even the second time I ran into it. I don't think there's a problem with choosing one default over the other, but if an alias is specified I think it could be safely assumed that casting is not necessary.

Is there any impact on nested relationships as well? Haven't had to use that feature yet.

dominikb:
Nested relationship names will be cast as well. (Can be seen in this [test case](https://github.com/spatie/laravel-query-builder/blob/1321c20de6a1f687b38963b289b0dd814b8beba5/tests/IncludeTest.php#L229))

I just found an older todo that refers to validating that included relationships do exist:
https://github.com/spatie/laravel-query-builder/blob/1321c20de6a1f687b38963b289b0dd814b8beba5/src/Concerns/AddsIncludesToQuery.php#L81
Could we remove camel casing and add this check to give the user a meaningful exception?

pionl:
Hi guys, I would recommend to be able to use a config entry to determine if CamelCase should be used?

AlexVanderbist:
I agree that this is not the best default behaviour. Initially, I only added this because I didn't like the look of case sensitive URL parameters 🙈. Is there any reason to keep this in? I'd accept a PR that just gets rid of this on the v3 branch.

I'd accept a PR for that non-existing relationship exception as well.

freekmurze:
We'll take a look at this in the next major version.

PR #794, opened 2022-07-27
Remove relation name conversion and respect included fields for relations

dominikb:
This PR addresses two issues:
- automatic snake_case conversion of included relations: #792, #440, #640
- respecting allowed fields for relationship fields: #681

I've tried to split them originally but the changes go hand-in-hand which is why I decided against it in the end.

Including a relation now requires specifying the exact relation name. This should clear up some confusion that arose with the automatic conversion in the past.

The key and foreign key columns are always queried and included automatically as they are required to associate the loaded models.
I don't know if this automatic resolution is something we want to support as I have a feeling that it won't always be possible.

```php
class TestModel extends Model
{
    public function relatedModels(): HasMany
    {
        return $this->hasMany(RelatedModel::class);
    }
}

// Previously
->allowedIncludes('relatedModels')
->allowedFields('related_models.name')

// Now
->allowedIncludes('relatedModels')
->allowedFields('relatedModels.name') //  Also automatically includes relatedModels.id
```

I've added a test case to verify that included fields for relations are respected, but I'm uncertain how the implementation holds up with more complex model setups. As I don't currently use this package in any of my projects, I can't test it extensively.

Please have a look at it and maybe test the changes. Any feedback or criticism is welcome.

### DIFF
--- a/src/Concerns/AddsFieldsToQuery.php
+++ b/src/Concerns/AddsFieldsToQuery.php
@@ -2,10 +2,10 @@
 
 namespace Spatie\QueryBuilder\Concerns;
 
-use ReflectionClass;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
-use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use ReflectionClass;
 use Spatie\QueryBuilder\Exceptions\AllowedFieldsMustBeCalledBeforeAllowedIncludes;
 use Spatie\QueryBuilder\Exceptions\InvalidFieldQuery;
 use Spatie\QueryBuilder\Exceptions\UnknownIncludedFieldsQuery;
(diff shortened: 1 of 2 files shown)


lrfahmi:
This pull request is pending?

leandrodiogenes:
Thank you so much @dominikb, this is exacly what im looking for.
Waiting the pull request aproval.

AlexVanderbist (approved):
LGTM for a 4.0 release. I'm also unsure if the logic in `resolveAdditionallyRequiredKeys` covers all relationships but I'm happy to move ahead with it if tests pass and the change is tagged as breaking.

VapoFred78:
Hello,
Will this pull request be merged one day?

andrewfloripa:
> Dear contributor,
> 
> because this pull request seems to be inactive for quite some time now, I've automatically closed it. If you feel this pull request deserves some attention from my human colleagues feel free to reopen it.

Regarding the correction reported by Dominikb about the plural.

Why was there no pull request?
Because I'm facing the same issue. I have a relationship with a singular name since it's a HasOne, and in the filter, I have to add an "S", forcing an incorrect plural outside of the Laravel convention.

leandrodiogenes:
+1

carvemerson:
Any chance to have this changes merged? @dominikb @AlexVanderbist

PR #818, opened 2022-10-26
Bugfix ignore allowed filters

davidjr82:
Currently [ignored filters values](https://spatie.be/docs/laravel-query-builder/v5/features/filtering#content-ignored-filters-values) option is working wrong because the function[ `resolveValueForFiltering` ](https://github.com/spatie/laravel-query-builder/blob/10f7869790e749d7d25f9c8ec4801068e955ced2/src/AllowedFilter.php#L145)is using `array_diff_assoc` that computes the difference of arrays **with additional index check** ([see PHP docs](https://www.php.net/manual/en/function.array-diff-assoc.php)).

This means that to ignore a filter, both value and key must match to be ignored. As filters are passed by GET params, they can be in a different order or have a non predictable index.

To ignore a value we need to match **only the value** without cheking the index, so php [`array_diff`](https://www.php.net/manual/en/function.array-diff.php) function is the right choice here.

I have modified an existing test to ignore the second option instead the first one to raise the error, and I have added one more test before commiting the bugfix.

### DIFF
--- a/src/AllowedFilter.php
+++ b/src/AllowedFilter.php
@@ -145,7 +145,7 @@ class AllowedFilter
     protected function resolveValueForFiltering($value)
     {
         if (is_array($value)) {
-            $remainingProperties = array_diff_assoc($value, $this->ignored->toArray());
+            $remainingProperties = array_diff($value, $this->ignored->toArray());
 
             return ! empty($remainingProperties) ? $remainingProperties : null;
         }


AlexVanderbist:
Awesome, can't believe this has gone unnoticed for so long. Thanks!

stevebauman:
Unfortunately this PR has broken a use-case for me in https://github.com/spatie/laravel-query-builder/issues/826.

Issue #797, opened 2022-08-02
Create from mixes static and self


olliescase:
https://github.com/spatie/laravel-query-builder/blob/9b7bf9123a66ddd38e0e2e9c1138b69c2325b11a/src/QueryBuilderRequest.php#L32

Is there a reason why the `fromRequest` method creates a new instance of `self()` rather than `static()` this means if you extend the class you have to then also override this function in order for it to return the extended class.

There may well be a "best practice" or use-case for this behaviour I'm not thinking of, but thought I'd question it as it struck me a little odd :)

AlexVanderbist:
Hey @olliescase, you're right, this is probably an oversight. Feel free to PR the change from `new self()` to `new static()` and I'll merge it in :)

Thanks!

PR #819, opened 2022-10-27
Change self to static when creating query builder

olliescase:
Fix for #797

fromRequest mixed static and self; this fixes to only use `static`


AlexVanderbist:
Awesome, thanks!

PR #821, opened 2022-10-31
Add BeginsWithStrict filter

### LINKED ISSUE #820: FiltersPartial index miss
FiltersPartial is using LOWER() which causes index misses. Many collations are case insensitive so LOWER() is not necessary.

The LIKE expression uses % before the search value which also cases index misses. This is necessary at times but if it could be used optionally that would be ideal.

Perhaps instead of just FiltersPartial, we add FiltersBeginsWith and FiltersEndsWith as options to place % only at the beginning or end.

danilopinotti:
## Introduction
The current "FiltersPartial" filter is very useful, but, it has some performance issues when we use it in a table with many data (see #820).

This PR introduces the "beginsWithStrict" filter, which may be used in situations described before.

In addition, this PR also allows us to extend the PartialFilter to change only the "where" clause applies (method `getWhereRawParameters`), making it possible implements other cases, like `beginsWith`, `endsWith`, `endsWithStrict`, etc.

## Using
```php
$users = QueryBuilder::for(User::with('media'))
    ->allowedFilters([
        AllowedFilter::beginsWithStrict('name'),
    ])
   // ...
```

## TODO
- [x] Implements BeginsWithStrict
- [x] Cover with Tests

## Related
Closes #820

### DIFF
--- a/src/AllowedFilter.php
+++ b/src/AllowedFilter.php
@@ -4,6 +4,7 @@ namespace Spatie\QueryBuilder;
 
 use Illuminate\Support\Collection;
 use Spatie\QueryBuilder\Filters\Filter;
+use Spatie\QueryBuilder\Filters\FiltersBeginsWithStrict;
 use Spatie\QueryBuilder\Filters\FiltersCallback;
 use Spatie\QueryBuilder\Filters\FiltersExact;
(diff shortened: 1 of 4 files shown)


oceanapplications:
Merging this would make my day :)

AlexVanderbist:
Good addition, thanks!

PR #827, opened 2022-11-28
Fix `array_diff_assoc` BC break in v5.1.0

### LINKED ISSUE #826: [v5.1.0] BC Break -- `array_diff_assoc` vs `array_diff` -- Array to string conversion
PHP Version: 8.0.25
Laravel Version: 9.41.0
Query Builder Version: 5.1.0

After updating from v5.0.3 to v5.1.0, all of my tests with this package broke, with the below exception:

<details>
<summary>Exception Screenshot</summary>
Screenshot 2022-11-28 at 12 43 42 PM
</details>

This is because several of my filters contain multi-dimensional associative arrays, and not just key-value pairs.

For further context, we've built a condition system with this package. Dumping `$value` here:

https://github.com/spatie/laravel-query-builder/blob/09d22513dc02417781499b6b9e9b89bea5e85242/src/AllowedFilter.php#L155-L159

Results in:

```php
array:1 [▼
  0 => array:2 [▼
    "attribute" => "name"
    "operator" => "equals"
    "value" => "John Doe"    
  ]
]
```

However, it appears `array_diff` does not support multi-dimensional arrays, breaking the value resolution function and throwing the above exception.

Let me know if you require further information. I'll work on a PR in the meantime, along with a test case.

stevebauman:
Closes #826 

Passing tests: https://github.com/stevebauman/laravel-query-builder/actions/runs/3567702432

### DIFF
--- a/src/AllowedFilter.php
+++ b/src/AllowedFilter.php
@@ -154,7 +154,8 @@ class AllowedFilter
     {
         if (is_array($value)) {
             $remainingProperties = array_map(
-                [$this, 'resolveValueForFiltering'], $value
+                [$this, 'resolveValueForFiltering'],
+                $value
             );
 
             return ! empty($remainingProperties) ? $remainingProperties : null;
--- a/src/AllowedFilter.php
+++ b/src/AllowedFilter.php
@@ -2,7 +2,6 @@
 
 namespace Spatie\QueryBuilder;
 
-use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Spatie\QueryBuilder\Filters\Filter;
 use Spatie\QueryBuilder\Filters\FiltersBeginsWithStrict;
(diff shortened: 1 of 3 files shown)


davidjr82:
hi @stevebauman,

this PR covers the multidimensional case my PR didn't cover, that's great!

My only doubt here is that the ignored values will apply to any of the nested values of the multidimensional array, right? For example, you can not ignore only "John Doe" in the key "value" of your example in #826 . 

Thinking on how to improve it, right now `ignore` is just a collection of values. If we add the concept of key/value we would be able to ignore only when the combination of both happens, that is useful for multidimensional arrays.

But anyway, your PR improves the behaviour that has the package right now, so looks very good IMO.

stevebauman:
Hey @davidjr82, thanks for reviewing the PR, much appreciated 🙏 

> My only doubt here is that the ignored values will apply to any of the nested values of the multidimensional array, right?

stevebauman:
Hey @davidjr82, thanks for reviewing the PR, much appreciated 🙏 

> My only doubt here is that the ignored values will apply to any of the nested values of the multidimensional array, right?

You're correct -- if any array value is contained in the `ignored` collection, then it will be filtered out.

I agree with your improvement. We could tackle that in a following PR so that the bug can be patched and then the additional feature could be placed into branch and release.

freekmurze:
Thank you!

PR #830, opened 2022-12-12
 Add config to disable InvalidSortQuery exception

bohemima:
Just like the disable InvalidFilterQuery exception we should be able to silently ignore sorting exception


AlexVanderbist:
Cool, thanks for the PR!

PR #841, opened 2023-01-20
Fix workflow badges in README

nelson6e65:
Fix invalid Badge parameters (https://github.com/badges/shields/issues/8671)


freekmurze:
Thanks!

PR #846, opened 2023-02-08
Update sorting.md

shukriYusof:
SortDirection under AllowedSort class can be used as chaining method to query builder


AlexVanderbist:
Thanks!

PR #865, opened 2023-04-30
Change the way AllowedSorts are set

luilliarcec:
It turns out that it is the only property that is not set when building the QueryBuilder object, it is only set if there is a request with sort key. I'm doing this PR because I'm making some classes to document my APIs that add an integration between this package and Laravel Scribe, but I can't get the value of AllowedSorts because it's not set unless the request brings the sort key.

### DIFF
--- a/src/Concerns/SortsQuery.php
+++ b/src/Concerns/SortsQuery.php
@@ -12,12 +12,6 @@ trait SortsQuery
 
     public function allowedSorts($sorts): static
     {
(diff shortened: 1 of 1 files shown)


freekmurze:
The tests are failing for this one, could you have a look?

luilliarcec:
Hi, @freekmurze 

The test runs completely and is passing, but on completion it ends with an error code.

image

AlexVanderbist:
A bit late on our end but thanks for the PR!

PR #872, opened 2023-06-15
Accepts string value for query param "fields"

ezra-obiwale:
In addition to accepting GET param `fields` like

```
fields[table1]=col1,col2&fields[table2]=col1,col2
```

this PR makes it possible to accept values like

```
fields=table1.col1,table1.col2,table2.col1,table2.col2
```

And if `table1` is the table for the target resource, then it can simply be like

```
fields=col1,col2,table2.col1,table2.col2
```


BBrey:
We would really love this, because now we actually need to fill in the table name like table_link_to_table when using the fields

PR #880, opened 2023-07-28
Laravel Spatie Query Builder - create OR condition for filters

taskinbirtan:
Hello 

I have added a functionality to group the search within multiple columns.
I have addressed the issue at stackoverflow. Which I was working on also, and I thought that I have found an elegant solution without breaking spatie's functionalities.

Here is the issue: 
https://stackoverflow.com/questions/68578125/laravel-spatie-query-builder-create-or-condition-for-filters

Kind regards 

Birtan Taskin, Software Engineer

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,20 @@ $users = QueryBuilder::for(User::class)
 // all `User`s that contain the string "John" in their name
 ```
 
+### Grouping Multiple Filters (OR)
+
+### Filter a query based on a request: `/users?filter[search.name]=Birtan&filter[search.surname]=Taskin`:
+
+```php
+use Spatie\QueryBuilder\QueryBuilder;
+
+$users = QueryBuilder::for(User::class)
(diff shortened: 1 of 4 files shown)


freekmurze:
Could you take a look at the failing tests?

AlexVanderbist:
Hi @taskinbirtan, nice work! Unfortunately we decided against merging this at this time. We try to keep this package as abstract as possible (following [JSON API spec](https://jsonapi.org/)) while keeping it easy to extend it with your own functionality. While I personally like the idea of being able to `OR` requested filters, I don't see it fitting in the package unless the JSON API spec adds a standardised solution.

Fortunately, it should be pretty easy at this point to extract the `SearchFilter` class from your PR to it's own package (and optionally extend the `QueryBuilder` to make the required changes to the `addFiltersToQuery` method.

Thanks again for your work on this!

PR #895, opened 2023-09-12
Allow nullable filter

enricodelazzari:


enricodelazzari:
@AlexVanderbist Any news on this? Thanks

AlexVanderbist:
Another great PR, thanks!

patrickrobrecht:
@enricodelazzari @AlexVanderbist When upgrading from Laravel Query Builder 5.5.0 to 5.6.0, we noticed that this PR introduces a small change which is not fully backwards compatible as we expected according to Semantic Versioning. See PR #902 for details and a bugfix restoring backwards compatibility.

patrickrobrecht:
As mentioned above, the change introduced by this PR is not fully backwards compatible. I've updated #902 to introduce a new method which provides the same behavior as `default(null)` did before merging this PR. Could you have a look on that.

PR #898, opened 2023-10-01
Fix broken LIKE escaping

GrahamCampbell:
Test failures are not related to my changes.


AlexVanderbist:
Good catch! I've got the tests fixed on main so I'll merge this and run the tests on there before tagging. Thanks!

PR #902, opened 2023-10-18
Unset default of AllowedFilter after it was set once, which was possible with default(null) before nullable filters where implemented in 5.6.0

patrickrobrecht:
The changes from #895 aren't backwards compatible.

`AllowedFilter::scope('test')->default('myDefaultValue')->default(null)` does not unset the default value anymore.

Background: We have a static `allowedFilters` method in a model, such that we can use the same set of `AllowedFilter`s in multiple controllers: 

In `TestModel`:
```php
public static function allowedFilters(): array
{
        return [
                AllowedFilter::scope('test')->default('myDefaultValue'),
                // ...
        ];
}
```

We can use the following in multiple controllers:
```php
$testModels = QueryBuilder::for(TestModel::class)
        ->allowedFilters(TestModel::allowedFilters())
```

Removing one of the default values from the array `allowedFilters()` when using the filter worked with Laravel Query Builder 5.5.0, but not with 5.6.0:
```php
$allowedFilters = [];
foreach (TestModel::allowedFilters() as $allowedFilter) {
        $allowedFilter->default(null);
        $allowedFilters[] = $allowedFilter;
}
$testModels = QueryBuilder::for(TestModel::class)
        ->allowedFilters($allowedFilters);
```

This PR introduces a new `unsetDefault()` method which could be used instead of the `default(null)` call in the example above. This behavior is not fully backwards compatible as one should be able to expect according to Semantic Versioning used by this project.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -9,7 +9,7 @@ jobs:
             fail-fast: false
             matrix:
                 os: [ubuntu-latest]
-                php: [8.2, 8.1, 8.0]
+                php: [8.3, 8.2, 8.1, 8.0]
                 laravel: [9.*, 10.*]
                 stability: [prefer-lowest, prefer-stable]
                 include:
--- a/src/AllowedFilter.php
+++ b/src/AllowedFilter.php
@@ -153,6 +153,7 @@ class AllowedFilter
         $this->default = $value;
 
         if (is_null($value)) {
+            $this->unsetDefault();
             $this->nullable(true);
         }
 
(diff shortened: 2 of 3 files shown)


patrickrobrecht:
I've updated https://github.com/spatie/laravel-query-builder/pull/902 to introduce a new method `unsetDefault()` which provides the same behavior as `default(null)` did before merging #895. Of course, there are tests for the new method as well.

@AlexVanderbist Could you have a look on that?

AlexVanderbist:
Hey, sorry about breaking backwards compatibility. This approach looks good to me, so I'll merge it in and add a disclaimer in the `UPGRADING.md`. Thanks!

PR #906, opened 2023-11-02
add option disable_invalid_includes_query_exception

dimzeta:
By default when we add a non desired include, we automatically have an `InvalidIncludeQuery`. The option `disable_invalid_includes_query_exception` allow us to disable it. 

When this option is enabled,  we remove the non included relationship and having only allowed relationships.


siarheipashkevich:
@AlexVanderbist please take a look at this PR

AlexVanderbist:
Awesome, thanks! (Sorry for the delay)

PR #918, opened 2024-01-14
[Docs] Update config file content

shdehnavi:
I've updated the config file content in the installation & setup document page


freekmurze:
Thanks!

PR #943, opened 2024-05-21
Fix ability to filter models by an array as filter value

inmula:
In commit 9e68f80 filtering models by an array as a filter value was broken. This is confirmed by the tests that are failing: https://github.com/spatie/laravel-query-builder/actions/runs/9029978604/job/24813361498

This PR restores the original behaviour while keeping PHPstan happy.


freekmurze:
Thanks!

PR #964, opened 2024-08-14
AllowedFilter should return static rather than self

kosarinin:
Hey, everyone!

Thanks for this fantastic project. We find it very useful in our projects.

In our project, we want to extend the `AllowedFilter` class, and it seems nothing prevents us from doing that with the only exception - all methods return `self` and not `static`.

Please consider merging these changes.

I appreciate your work on this project!


Nielsvanpach:
Thanks!

PR #972, opened 2024-09-23
Refactor AddsIncludesToQuery trait to add default includes handling

inmass:
### Add defaultIncludes functionality to AddsIncludesToQuery trait

- Implement new defaultIncludes method to support automatic inclusion of relationships
- Add support for default include of 'primaryImage' relationship

This commit introduces the ability to specify default includes in the query builder,
which was not previously available. The new functionality allows certain relationships
to be automatically included in every query without explicitly requesting them.

Key changes:
1. New defaultIncludes method added to set default includes
2. Updated findInclude method to consider both allowed and default includes
3. Modified addIncludesToQuery to handle default includes

Example usage:
`$query->defaultIncludes(['primaryImage'])`

This addition enhances the flexibility of the query builder by allowing
automatic inclusion of critical relationships, improving ease of use and
reducing the need for repetitive include requests.

### DIFF
--- a/src/Concerns/AddsIncludesToQuery.php
+++ b/src/Concerns/AddsIncludesToQuery.php
@@ -11,12 +11,35 @@ use Spatie\QueryBuilder\Includes\IncludeInterface;
 trait AddsIncludesToQuery
 {
     protected ?Collection $allowedIncludes = null;
+    protected ?Collection $defaultIncludes = null;
 
     public function allowedIncludes($includes): static
     {
         $includes = is_array($includes) ? $includes : func_get_args();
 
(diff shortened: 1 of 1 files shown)


AlexVanderbist:
Hi @inmass, thanks for your PR! How does this differ from calling `->with('related')` on the query builder object? So for example:

```php
// This will always load in the `posts` relationship, regardless of the request URL
QueryBuilder::for(User::class)->with('posts')->allowedIncludes(['comments'])->get();
```

inmass:
@AlexVanderbist  Thank you for your comment! apparently i missed that one, my PR is basically trying to reach the same result as using `with()`

Issue #959, opened 2024-07-19
Selecting fields for included relations


JaviLerma:
Greetings from Argentina!
The [Official docs](https://spatie.be/docs/laravel-query-builder/v5/features/selecting-fields#content-selecting-fields-for-included-relations) says:

```php
QueryBuilder::for(Post::class)  
    ->allowedFields('author.id', 'author.name')  
    ->allowedIncludes('author')  
    ->get();
```

`GET /posts?include=author&fields[author]=id,name`

Return all post but only _id_ and _name_ of the _author_, but it doesn't work. It fetches _all_ the authors fields.

Relationships

User model:

```php
public function posts(): HasMany
    {
        return $this->hasMany(Post::class, 'author_id');
    }}
```
    
 Post model:
 
 ```php
 public function author(): BelongsTo
    {
        return $this->belongsTo(User::class, 'author_id');
    }
```

I am doing something wrong or its a bug? 
Thanks

luizmagao:
Estou passando pelo mesmo problema, não sei como resolver ainda.

inmass:
The current implementation doesn't work as expected due to several factors related to naming conventions and configuration:

1. Relationship and Table Naming:
By default, the query builder expects plural names for relations and table names. For example:
   - 'author' should be 'authors'
   - 'postComment' should be 'post_comments'

2. Using Singular Names:
If you prefer to use singular naming for relationships, you need to publish the query-builder config file:
  `php artisan vendor:publish --provider="Spatie\QueryBuilder\QueryBuilderServiceProvider" --tag="query-builder-config"`
Then, set `convert_relation_names_to_snake_case_plural` to `false` in the config file.

3. Relation ID Fields:
Ensure you add the relation id to the fields param

4. Query Syntax:
You can query related fields using either of these syntaxes:
- Dot notation: `fields=id,title,car_model_id,carModel.id,carModel.slug`
- Array notation: `fields[carModel]=id,slug`

These adjustments should resolve the issues with relationship querying and naming conventions in the current implementation.

PR #973, opened 2024-09-23
feat: Automatically include 'id' field in related table queries

inmass:
(this may be also related to issue #959
# Automatically Include 'id' Field in Related Table Queries

## Description
This PR modifies the `getRequestedFieldsForRelatedTable` method to automatically include the 'id' field when querying related tables. This enhancement improves user experience by reducing the need for redundant specifications in API requests.

## Motivation and Context
When working with related tables, the 'id' field is often crucial for maintaining relationships and performing subsequent operations. However, requiring users to explicitly include 'id' in every request (e.g., `fields[carModel]=slug,id`) can be cumbersome and error-prone. This change aims to streamline the API usage while ensuring that essential data is always available.

### DIFF
--- a/docs/features/selecting-fields.md
+++ b/docs/features/selecting-fields.md
@@ -42,13 +42,24 @@ $users = QueryBuilder::for(User::class)
 Selecting fields for included models works the same way. This is especially useful when you only need a couple of columns from an included relationship. Consider the following example:
 
 ```php
-GET /posts?include=author&fields[author]=id,name
+GET /posts?include=author&fields[authors]=name
 
 QueryBuilder::for(Post::class)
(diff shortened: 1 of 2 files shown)


AlexVanderbist:
Hi @inmass, thanks for another interesting contribution! I agree that it's not the best DX to include the relevant `id` columns on included fields all the time. However, I never found a good fix for this because there's a bunch of edge cases where it's not as simple as adding a select on `id`. For example, some people use `uuid` as a key, while others use custom foreign key columns like `related_id`. There are also special relationship types, such as polymorphic relations, which require more than one column to clarify the relationship. Maybe I see the issue as more complicated than it actually is, but let's get some test cases going first to see where this breaks :)

PR #975, opened 2024-10-07
New Filter belongs to

gpibarra:
This new filter receives the id and allows to call the native method `whereBelongsTo` method of the model ([doc](https://laravel.com/docs/11.x/eloquent-relationships#querying-belongs-to-relationships)).
The method accepts nested relationships.
The filter accepts receiving an array with the ids.

Unfortunately my English is not good and I couldn't write the documentation correctly.
If you think the PR is correct, could you help me with the documentation?


src/Filters/FiltersBelongsTo.php
```diff
@@ -0,0 +1,96 @@
...
+
+        $relatedModel = $relationObject->getRelated();
+        if (! is_subclass_of($relatedModel, Model::class)) {
+            return null;
+        }
+
+        return $relatedModel;
+    }
+
+    protected function getModelFromRelation(Model $model, string $relation, int $level = 0): ?Model
```
AlexVanderbist, line 81: Is `$level` still being used? I can see it's passed on recursively but never used.

src/Exceptions/InvalidFilterProperty.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Spatie\QueryBuilder\Exceptions;
+
+use Exception;
+
+class InvalidFilterProperty extends Exception
+{
+    public static function make($property): static
+    {
+        return new static("Filter property `{$property}` is invalid.");
```
AlexVanderbist, line 11: Can we add a `$reason` to this exception? From the implementation I see there's only one reason this happens: `belongs to relationship is not an Eloquent relationship`.

AlexVanderbist (changes requested):
Hi, thanks for the PR! Two small review comments, after that I'm happy to tweak the docs and merge this :)

src/Filters/FiltersBelongsTo.php
gpibarra: As a habit, whenever I make a recursive function I send the level parameter, sometimes it is necessary to determine the first iteration of the rest. In this case it was not necessary and I forgot to delete it. You can delete it safely.

src/Exceptions/InvalidFilterProperty.php
gpibarra: Exactly.
There was another exception that I wasn't handling (on purpose), when the relation (or the last relation in a nested dotted string) is not a BelongTo relation, in that case an unhandled exception was thrown but Laravel gave a correct message and context to find this error. This case was not covered in the tests.
Based on that exception, now remove `InvalidFilterProperty` and throw that same exception `RelationNotFoundException`. I also added more test cases that cover all the options (this case and the previous one).
This native Laravel exception gives a more specific message to fix the error.

gpibarra: @AlexVanderbist Have you been able to review this?

gpibarra: @freekmurze Have you been able to review this?

AlexVanderbist: Perfect, thank you!

gpibarra:
Thanks @AlexVanderbist! The documentation is pending.
[https://github.com/spatie/laravel-query-builder/blob/main/docs/features/filtering.md#belongsto-filters](https://github.com/spatie/laravel-query-builder/blob/main/docs/features/filtering.md#belongsto-filters)

PR #978, opened 2024-10-19
Fixed IncludedCount.php

dash8x:
Fixed bug where the include count fails if a relation has the word `Count` in it. For example, it fails for `billingCountry` by removing the word `Country` and trying to load a `billing` relation.


freekmurze:
Thanks!

PR #983, opened 2024-12-04
feat: implement a way of matching the fields to the jsonapi spec

CoolGoose:
Hello folks, this is just a PoC (that I am cleaning up as we speak) of figuring out if there's interest on another config option to allow  mapping closer to the JsonAPI spec and making this easier to use from the start with https://github.com/timacdonald/json-api

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -37,6 +37,7 @@ class TestCase extends Orchestra
             $table->increments('id');
             $table->timestamps();
             $table->string('name')->nullable();
+            $table->string('full_name')->nullable();
             $table->double('salary')->nullable();
             $table->boolean('is_visible')->default(true);
         });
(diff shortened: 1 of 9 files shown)


CoolGoose:
I think it's ready for a review 😅

tests/TestCase.php
```diff
@@ -92,7 +94,7 @@ protected function setUpDatabase(Application $app)
     protected function getPackageProviders($app)
     {
         return [
-            RayServiceProvider::class,
+            // RayServiceProvider::class,
```
pedro0125, line 97: Why do we need to comment this?

CoolGoose: Tests don't work otherwise. I am not sure who introduced that service provider, but it's not in the composer dependencies list.

utils/run_tests.sh
AlexVanderbist, line 1: Gonna remove this one after merging

AlexVanderbist:
Hey @CoolGoose, thank you for taking the time to make this more compatible with the base JSON API specification and Timacdonald's package. Although it adds some complexity, I'm happy to merge it since the tests are passing and no breaking changes have been introduced.

utils/run_tests.sh
CoolGoose: thank you for the review. As a curiosity @AlexVanderbist how does one run the tests ? I'm happy to have a separate PR for this if it helps, or at least learn how you guys run the tests internally.

PR #986, opened 2024-12-23
Fix selecting fields on belongs to many relations

rasmuscnielsen:
When using field selection on belongs-to-many relationships, it currently throws an "id in field list is ambiguous" error:

> Integrity constraint violation: 1052 Column 'id' in field list is ambiguous (Connection: mysql, SQL: select `id`, `name`, `pivot_models`.`test_model_id` as `pivot_test_model_id`, `pivot_models`.`related_through_pivot_model_id` as `pivot_related_through_pivot_model_id` from `related_through_pivot_models` inner join `pivot_models` on `related_through_pivot_models`.`id` = `pivot_models`.`related_through_pivot_model_id` where `pivot_models`.`test_model_id` in (1))

This PR fixes that by qualifying relationship fields upon select.

### DIFF
--- a/src/Includes/IncludedRelationship.php
+++ b/src/Includes/IncludedRelationship.php
@@ -28,7 +28,7 @@ class IncludedRelationship implements IncludeInterface
                 }
 
                 return [$fullRelationName => function ($query) use ($fields) {
-                    $query->select($fields);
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #987, opened 2024-12-24
feat: add conditionable allowed filter

luilliarcec:
This adds the possibility of providing a closure to the `AllowedFilter` class, to determine whether or not it should be applied.

### DIFF
--- a/src/AllowedFilter.php
+++ b/src/AllowedFilter.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\QueryBuilder;
 
+use Closure;
 use Illuminate\Support\Collection;
 use Spatie\QueryBuilder\Enums\FilterOperator;
 use Spatie\QueryBuilder\Filters\Filter;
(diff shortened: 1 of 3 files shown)


luilliarcec:
> Hey @luilliarcec, thanks for the PR!
> 
> Currently, it's already possible to dynamically "disable" filters in the `allowedFilters` method by simply omitting them.
> 
> For example:
> 
> ```
> $queryBuilder
>     ->allowedFilters(array_filter([
>         $someCondition ? AllowedFilter::exact('user_id') : null
>     ]))
>     ->get();
> ```
> 
> I agree that the `->when` syntax is more fluent tho, but I'm curious if you have a practical applications for this?

Hi @AlexVanderbist 

I don't want to pass up the opportunity to wish you and the spatie team a happy new year, thanks for your great contributions to the Laravel community.

Well, I was born the need, when in an application, certain users could make use of a filter in a matter of a role or permission.

PR #991, opened 2025-01-21
support nulls in belongsTo

gpibarra:
This is an improvement on PR #975.

The initial motivation is to support the search for belongsTo=null.

In these cases, it is necessary that the database, the column allows nulls, for example `$table->foreignId('user_id')->nullable()->constrained(table: 'users', indexName: 'posts_user_id');`.

In cases where it is detected that a nullable value is sent (`null`, `'null'`, `0`, `'0'`, `''`), then `whereNull($column)` is added to the query.

In this PR the code was simplified, instead of using `whereBelongsTo($collection, $relation)`, `whereIn($relation->getQualifiedForeignKeyName(), $values)` was used; where `$values` is no longer a collection but an array (with the ids).

In cases where a nullable value and an id are sent at the same time, for example `[null, 1, 3]`, both wheres are executed using the `OR` operator:
`$query->where(fn () => $q->orWhereNull($relationColumn)->orWhereIn($relationColumn, $values)`

---

As I said in #975 

> Unfortunately my English is not good and I couldn't write the documentation correctly.
If you think the PR is correct, could you help me with the documentation?

The documentation is pending.
[https://github.com/spatie/laravel-query-builder/blob/main/docs/features/filtering.md#belongsto-filters](https://github.com/spatie/laravel-query-builder/blob/main/docs/features/filtering.md#belongsto-filters)

### DIFF
--- a/src/Filters/FiltersBelongsTo.php
+++ b/src/Filters/FiltersBelongsTo.php
@@ -51,7 +51,7 @@ class FiltersBelongsTo implements Filter
     {
         return array_filter(
             $values,
-            fn ($v) => ! in_array($v, [null, 0, 'null', '0'], true)
+            fn ($v) => ! in_array($v, [null, 0, 'null', '0', ''], true)
         );
     }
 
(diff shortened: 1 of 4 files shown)


gpibarra:
@AlexVanderbist Have you had a chance to review this?

PR #993, opened 2025-02-18
Exclude `.github` folder and `.php_cs` from being included in composer installation

stevebauman:
This PR excludes the `.github` folder and the `.php_cs.dist.php` file from the composer installation:

Screenshot 2025-02-18 at 5 27 48 PM


freekmurze:
Thanks!

PR #1002, opened 2025-04-15
Enhance QueryBuilder with generics support for better type inference

alexkart:
This pull request includes several changes to the `QueryBuilder` class in `src/QueryBuilder.php` to improve type safety and code clarity by adding template annotations and type hints. The most important changes include adding a template annotation to the class, updating the return type annotations for methods, and ensuring proper type casting.

When trying to specify the correct type in the following code:

```php
/**
     * @return LengthAwarePaginator<int, Client>
     */
    public function queryClients(Request $request): LengthAwarePaginator
    {
        /** @var QueryBuilder<Client> $queryBuilder */
        $queryBuilder = QueryBuilder::for(Client::class, $request);
...
```
Static analysis (PHPStan) generates the following error:
```
Method Modules\Client\Repositories\ClientRepository::queryClients() should return Illuminate\Pagination\LengthAwarePaginator<int, Modules\Client\Models\Client> but returns  
         Illuminate\Pagination\LengthAwarePaginator<int, Illuminate\Database\Eloquent\Model>.                                                                                         
         🪪  return.type
```
It happens because `Illuminate\Database\Eloquent\Builder` is a generic class:
```php
/**
 * @template TModel of \Illuminate\Database\Eloquent\Model
 */
class Builder implements BuilderContract
```

but `\Spatie\QueryBuilder\QueryBuilder` that specifies this class as a `@mixin` is not generic, so there is no way to specify a concrete type for the `TModel`.
This PR adds the ability to do this.

### DIFF
--- a/src/QueryBuilder.php
+++ b/src/QueryBuilder.php
@@ -14,7 +14,8 @@ use Spatie\QueryBuilder\Concerns\FiltersQuery;
 use Spatie\QueryBuilder\Concerns\SortsQuery;
 
 /**
- * @mixin EloquentBuilder
+ * @template TModel of Model
+ * @mixin EloquentBuilder<TModel>
  */
 class QueryBuilder implements ArrayAccess
 {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #1008, opened 2025-05-19
Docs: clarify selecting fields for included relations

amyavari:
Based on Issue #959 and some other misunderstandings included in the docs, I tried to make it clearer for users to understand the deeper behavior of Spatie Query Builder when selecting fields for relationships.

- Updated the `Installation & setup` page with all available configuration options from the package config file.
- Added a note about default name convention for the `Selecting fields for included relations` section in the `Selecting fields` page, and fixed its example.
- Updated the example in the explanation for the `convert_relation_names_to_snake_case_plural` option.
- Updated the explanation for the `convert_relation_table_name_strategy` option to clarify its underlying behavior and purpose.
- Fixed outdated references in the `Selecting included fields` section on the `Including relationships` page.

### DIFF
--- a/config/query-builder.php
+++ b/config/query-builder.php
@@ -62,12 +62,14 @@ return [
     'convert_relation_names_to_snake_case_plural' => true,
 
     /*
-     * By default, the package expects relationship names to be snake case plural when using fields[relationship].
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks!

PR #1010, opened 2025-07-04
Allow nested filters

bambamboole:
This PR adds the option to pass multiple arrays as filters. 
Just a small convenience which is also implemented for includes and fields


freekmurze:
Thanks!

PR #1059, opened 2026-04-30
docs: add Angular front-end implementation

AndreaAlhena:
Angular: [ng-qubee](https://www.npmjs.com/package/ng-qubee) by [AndreaAlhena](https://github.com/AndreaAlhena) an Angular library to fluently build query strings compatible with this package, supporting filters, sorts, includes, pagination, appends and fields.


freekmurze:
Thank you!

